PR #110, opened 2024-10-11
cli: Improve error message if token is missing

aborg-dev:
Ran into this while debugging missing ZISK_TOKEN issue


aborg-dev:
Actually, this is not relevant after rebase, so closing.

PR #112, opened 2024-10-14
New version of range check

hecmas:


xavi-pinsach (commented):
this has already included in https://github.com/0xPolygonHermez/zisk/pull/122

PR #130, opened 2024-10-22
Separating shift and sign extend

hecmas:


hecmas:
Closed because, even tho de PIL is cleaner, the numbers do not pay off.

PR #163, opened 2024-11-14
verify_all.sh first drafts, and fixes

fractasy:


fractasy:
Replaced by another PR

PR #165, opened 2024-11-18
Verify all + riscof tests fixes

fractasy:


core/src/zisk_ops.rs
```diff
@@ -823,9 +823,9 @@ pub fn opc_max(ctx: &mut InstContext) {
 #[inline(always)]
 pub const fn op_maxu_w(a: u64, b: u64) -> (u64, bool) {
     if (a as u32) > (b as u32) {
-        (a, false)
+        (a as i32 as i64 as u64, false)
```
sam0x17, line 826: this seems rather indirect, can you go direct from `u32` to `u64`?

fractasy: The change was made because as per RISC-V spec atomic word operations must be extended as if they were signed, i.e. padding with F's if 31st bit is a 1.

fractasy: Currently, max & min opcodes are only called from the atomic RISC-V opcodes.  If we ever have to support other executable formats, we might have to create new max & min functions that behave the intuitive way.

PR #172, opened 2024-11-23
Documentation improvements

fractasy:


core/src/zisk_definitions.rs
```diff
@@ -156,3 +155,40 @@ pub const REG_T3: u64 = REG_X28; // Temporary register 3
...
+pub const P2_22: u64 = 0x400000;
+pub const P2_23: u64 = 0x800000;
+pub const P2_24: u64 = 0x1000000;
+pub const P2_25: u64 = 0x2000000;
+pub const P2_26: u64 = 0x4000000;
+pub const P2_27: u64 = 0x8000000;
+pub const P2_28: u64 = 0x10000000;
+pub const P2_29: u64 = 0x20000000;
+pub const P2_30: u64 = 0x40000000;
+pub const P2_31: u64 = 0x80000000;
```
sam0x17, line 191: this would be a really good application for a const fn that just calculates the power of 2 number. Since it would be a const fn, it would execute at compile-time, removing the need for the constants

sam0x17 (approved):
looks good but see comment about adding a const fn instead of power of 2 constants

core/src/zisk_definitions.rs
fractasy: Considered as a future improvement

PR #211, opened 2025-01-24
feature/new arch pubout

zkronos73:


ziskos/entrypoint/src/lib.rs
RogerTaule, line 1: Shouldn't the number of maximum public outputs be 32?

state-machines/mem/src/mem_planner.rs
```diff
@@ -1,6 +1,14 @@
 use std::sync::Arc;
 
 use sm_common::{BusDeviceMetrics, ChunkId, Plan, Planner};
+#[cfg(feature = "debug_mem")]
+use crate::MemDebug;
+#[cfg(feature = "debug_mem")]
+use zisk_pil::MEM_ALIGN_AIR_IDS;
+
+#[cfg(feature = "debug_mem")]
+use crate::{MemAlignCheckPoint, MemHelpers};
```
RogerTaule, line 10: This crates are not used. Same with MEM_ALIGN_AIR_IDS. On the other hand, MEM_BYTES should be imported or otherwise will not compile with features debug_mem

RogerTaule (changes requested):

ziskos/entrypoint/src/lib.rs
zkronos73: Really, are 32 of 64 bits, in this point of code are 64 of 32 bits.

state-machines/mem/src/mem_planner.rs
zkronos73: I fixed it, removing no used parts, and use MemHelpers to get addr from addr_w

PR #268, opened 2025-02-23
data bus improvement changing Vec<_> to &[_] to avoid use to_vec() on…

xavi-pinsach:
… every databus write from emu


zkronos73 (approved):
:smile:

PR #480, opened 2025-08-06
fix: Add tests for edge cases discovered from golang

kevaundray:
To run locally:

```bash
cargo build --bin cargo-zisk --bin ziskemu
cd elf-regressions
./scripts/build.sh
./scripts/test.sh rom-setup all # Runs rom-setup on all elf files
 ./scripts/test.sh emu all # Runs ziskemu on all elf files
```


elf-regressions/jalr/test.s
```diff
@@ -0,0 +1,18 @@
+.section .text.init
```
kevaundray, line 1: This passes and is the same program in jalr_custom -- its here to show that changing the entrypoint, but keeping the same ROM causes rom-setup to fail

elf-regressions/fips_markers/test.s
```diff
@@ -0,0 +1,28 @@
+# Test: FIPS markers with entry point between them
+# Simulates Go binaries where _start (the code) is between FIPS boundary markers
+
+.section .text
+
+# FIPS start marker (data, not code)
+.global _fipsstart
+_fipsstart:
+    .word 0x46495053  # "FIPS"
+    .word 0x53544152  # "STAR"
```
kevaundray, line 10: I need to check if these are usually aligned -- read that its a single byte

elf-regressions/divw/test.s
```diff
@@ -0,0 +1,16 @@
+# Test case for DivW (32-bit signed division) bug
```
kevaundray, line 1: Fixed by https://github.com/0xPolygonHermez/zisk/pull/473

elf-regressions/remainder/remuw_op.s
```diff
@@ -0,0 +1,19 @@
+# Test case for RemuW (32-bit unsigned remainder) formatting bug
```
kevaundray, line 1: Fixed by https://github.com/0xPolygonHermez/zisk/pull/473

PR #481, opened 2025-08-07
fix: Compute low address in the jump dest table using the entrypoint to determine the boundary

kevaundray:
Related to #480

### DIFF
--- a/core/src/zisk_rom.rs
+++ b/core/src/zisk_rom.rs
@@ -114,6 +114,10 @@ pub struct ZiskRom {
     /// List of instruction program counter (address) in incremental order:
     /// 0x1000, 0x1004, ..., 0x80000000, 0x80000004, ...
     pub sorted_pc_list: Vec<u64>,
+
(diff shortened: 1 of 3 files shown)


core/src/zisk_rom.rs
```diff
@@ -114,6 +114,10 @@ pub struct ZiskRom {
     /// List of instruction program counter (address) in incremental order:
     /// 0x1000, 0x1004, ..., 0x80000000, 0x80000004, ...
     pub sorted_pc_list: Vec<u64>,
+
+    /// Minimum rom instruction PC (first program instruction address)
+    /// This is typically 0x80000000 but can be different (e.g., 0x80001000 with Go's internal linker)
+    pub min_program_pc: u64,
 }
```
kevaundray, line 121: Should also change this in get_instruction, get_mut_instruction -- currently it uses ROM_ADDR

kevaundray:
Fixing as this is handled in #486

PR #482, opened 2025-08-07
fix: handle nop instruction

kevaundray:
The initial zero instruction test case in #480 is exercising two things, so leaving this in draft for now


kevaundray:
Fixing as this is handled in #486

PR #484, opened 2025-08-07
feat: Modularize `elf2rom` and add tests

kevaundray:
Most of the LOC are adding tests for the functionality that has been outlined

### DIFF
--- a/core/src/elf_extraction.rs
+++ b/core/src/elf_extraction.rs
@@ -1,0 +1,283 @@
+//! ELF file extraction utilities for separating ELF parsing from ZiskRom population
+
+use elf::{
+    abi::{SHF_ALLOC, SHF_EXECINSTR, SHF_WRITE, SHT_PROGBITS},
+    endian::AnyEndian,
+    ElfBytes,
+};
+use std::{error::Error, fs, path::Path};
+
+use crate::{is_elf_file, RAM_ADDR, RAM_SIZE};
+
+const RAM_START_ADDR: u64 = RAM_ADDR;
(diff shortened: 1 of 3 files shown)


core/src/elf_extraction.rs
```diff
@@ -0,0 +1,284 @@
...
+pub fn merge_adjacent_ro_sections(sections: &[DataSection]) -> Vec<DataSection> {
+    if sections.is_empty() {
+        return Vec::new();
+    }
+
+    let mut merged = Vec::new();
+    let mut sections = sections.to_vec();
+
+    // Sort by address
+    sections.sort_by_key(|s| s.addr);
```
kevaundray, line 131: Doing the sorting here, once we have all of the sections means:

- This is O(n log n)
- If the sections are out of order, we still don't miss any

core/src/elf_extraction.rs
```diff
@@ -0,0 +1,283 @@
...
+use crate::{is_elf_file, RAM_ADDR, RAM_SIZE};
+
+const RAM_START_ADDR: u64 = RAM_ADDR;
+const RAM_END_ADDR: u64 = RAM_ADDR + RAM_SIZE;
+
+/// Raw bytes of `data` that will live at `addr` once the ROM has booted.
+#[derive(Debug, Clone)]
+pub struct DataSection {
+    pub addr: u64,
+    pub data: Vec<u8>,
```
kevaundray, line 19: This includes code, so maybe best to rename it to RawSection, so we don't get confused with Data as in .rodata

kevaundray:
This is done in #486

PR #485, opened 2025-08-07
chore: use `chunks_exact` in `convert_vector`

kevaundray:
Uses chunks_exact to avoid the indexing and bounds checks


zkronos73:
@cla-bot recheck

PR #541, opened 2025-09-09
feat: modular riscv decoder

kevaundray:


riscv_decoder/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+            bytes.len().is_multiple_of(expected_code_alignment),
+            "code length = {} which is not a multiple of {}",
+            bytes.len(),
+            expected_code_alignment
+        );
+
+        let mut instructions = Vec::with_capacity(bytes.len() / 2);
+        let mut i = 0;
+
+        while i + 2 <= bytes.len() {
```
kevaundray, line 47: This number 2 is the code_alignment or IALIGN to reference the specs

fractasy:
We stay with the current riscv decoder.  Thank you, Kev!

PR #589, opened 2025-10-13
added new precompile add256

zkronos73:
Added a new generic pil for bigint addition, with carry_in and carry_out, instanced for 256 bits with this interface:
```rust
pub struct SyscallAdd256Params<'a> {
    pub a: &'a [u64; 4],
    pub b: &'a [u64; 4],
    pub cin: u64,
    pub c: &'a mut [u64; 4],
}

pub extern "C" fn syscall_add256(params: &mut SyscallAdd256Params) -> u64 
```
Added option to return a value for precompiled without using memory, in this case  the carry out.


hecmas (approved):
LGTM

PR #594, opened 2025-10-16
Binary extension audit

hecmas:


hecmas:
Implemented in PR #613

PR #601, opened 2025-10-21
update pils with bits column feature to avoid use a manual hint

zkronos73:
Update pils to remove manual hint witness_bits, now with new version of pil2-compiler (v0.8.0) this hint is auto generated when use bits column feature.


xavi-pinsach (approved):
LGTM

PR #631, opened 2025-10-29
Feature/input rom data wo intermediates

zkronos73:
- Reduce memory used by memory count and plan
- To avoid a large number of intermediate reads, use a range check to demostrate increments of addr on rom_data and input_data
- Add a proofval to disable rom_data when it isn't used, this feature is usefull for tests.


fractasy (approved):

hecmas (approved):
LGTM

PR #665, opened 2025-11-15
fix: correct ELF st_info type check

VolodymyrBg:
Correct the ELF symbol type detection by masking the low four bits of st_info (the type field) and comparing to STT_FUNC, per the ELF specification (ELF*_ST_TYPE(info) = info & 0x0f). The previous condition used a bitwise-and test against STT_FUNC, which is a value, not a mask, and could allow false positives such as TLS or IFUNC symbols. We explicitly include STT_GNU_IFUNC to capture resolver functions that reside in code sections and should be treated as executable symbols. No other behavior or filtering was changed; we still gate on SymbolKind::Text to ensure code symbols.


kevaundray (commented):
LGTM!

VolodymyrBg:
> We require contributors/corporates @VolodymyrBg to read our Contributor License Agreement, please check the [Individual CLA document](https://github.com/0xPolygonHermez/clabot-config/blob/master/CLA%20Zisk.pdf)

Thanks, checked, bot not sure how to sign it to pass CI?

PR #686, opened 2025-12-17
Fix/ziskemu stats callstack jal

zkronos73:
Fix a problem with call using static jal, on zisk no active store_pc flag because use static compilation value of jmp_offset.
Fix a problem on stats when exists a non-used symbol main.
Add more friendly information when stack mismatch.


hecmas (approved):
LGTM

PR #689, opened 2025-12-18
Adding input path

hecmas:


hecmas:
This will be already included in 0.16.0

PR #701, opened 2026-01-07
Testing Fusaka blocks

hecmas:
Rerun tests when #697 is merged.


hecmas:
We decided that testing blocks for an Ethereum client is not the most appropriate tests for the ZisK CI.

PR #731, opened 2026-01-28
Fix/mtrace realloc

zkronos73:
Fix some bugs:
- Update the assembly runner (client) to support dynamic realloc of trace in the middle of chunk. 
- Update the register to obtain STEPS useds in last verification before realloc trace.
Add conditional compilation for diferent trace targets
Add some logs to detect trace reallocation


RogerTaule (approved):
LGTM

PR #787, opened 2026-02-13
Add documentation for hints generation from guest

agnusmor:


book/getting_started/precompile_hints.md
```diff
@@ -249,3 +261,142 @@ let processor = HintsProcessor::builder(my_sink)
...
+}
+```
+
+This call generates the hint input data using the exact input values that will later be used by the ZisK zkVM when executing the `zkvm/zisk` target code. This hint input data is consumed later by the `HintsProcessor`, allowing the SHA-256 computation to be performed outside the zkVM while remaining fully verifiable inside the circuit.
+
+After the hint generation, execution continues in the native target code to compute the SHA-256 result.
+
+From the guest program, we generate hints containing the input data for the corresponding `zisklib` functions (in this example, the `sha256_c` function). These `zisklib` functions may internally invoke one or more precompiles to produce the final result.
+
+When the hints are processed by the `HintsProcessor`, it executes the same ``zisklib function using the implementation code for the zkvm/zisk target. This produces the exact precompile results expected when executing the guest ELF inside the zkVM.
```
xavi-pinsach:
```suggestion
When the hints are processed by the `HintsProcessor`, it executes the same `zisklib` function using the implementation code for the zkvm/zisk target. This produces the exact precompile results expected when executing the guest ELF inside the zkVM.
```

xavi-pinsach (approved):

Issue #793, opened 2026-02-17
Consider replacing custom rust fork with rv64imac-unknown-none-elf


kevaundray:
## Description

Rust doesn't have out of the box compiler support for rv64ima, so there is currently a Rust fork that is used to support this. Now that ZisK has added support for compressed instructions, it is possible to remove the rust fork and use `rv64imac-unknown-none-elf` directly.

The main advantage being that you won't need to support a fork.

One disadvantage here is that rv64imac-unknown-none-elf is `no_std`, which may require some changes to make it work.

I've scoped out the rough changes needed [here](https://github.com/0xPolygonHermez/zisk/pull/804)

PR #801, opened 2026-02-18
chore(no_std): make `sys_rand` related code `no_std`

kevaundray:

In order to support rv64imac-unknown-none-elf or rv64im-unknown-none-elf in the future, ZiskOS would need to be no_std compatible. The sys_rand related code was using StdRng and the Mutex from the standard lib.

### DIFF
--- a/ziskos/entrypoint/Cargo.toml
+++ b/ziskos/entrypoint/Cargo.toml
@@ -17,7 +17,8 @@ num-traits = { workspace = true }
 precompiles-helpers = { workspace = true }
 
 lazy_static = "1.5.0"
-rand = "0.8.5"
+spin = "0.9"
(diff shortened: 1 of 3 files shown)


ziskos/entrypoint/Cargo.toml
```diff
@@ -17,7 +17,8 @@ num-traits = { workspace = true }
 precompiles-helpers = { workspace = true }
 
 lazy_static = "1.5.0"
-rand = "0.8.5"
+spin = "0.9"
+rand = { version = "0.8.5", default-features = false, features = ["small_rng"] }
```
kevaundray, line 21: small_rng is `no_std` -- note that StdRng, what is code was previously using and SmallRng make no portability guarantees as mentioned in the docs

hecmas:
Superseded by #965.

PR #802, opened 2026-02-18
chore(no_std): Replace `cfg.target` with `target.zisk_guest`

kevaundray:

In order to support rv64imac-unknown-none-elf or rv64im-unknown-none-elf in the future, ZiskOS would need to be no_std compatible -- This PR does some preparation to make it easier to migrate:

- cfg flags on `zisk_guest` and `zisk_host`: This makes it easier to still support rv64ima-unknown-zisk-elf. This forked rust compiler supports things like the std lib. In general, I think cfg flags on a feature is a bit cleaner in the source code. It is possible to have a build.rs that sets zisk_guest based off of target which would make it similar to what was happening before.
- Explicitly pass zisk_guest or zisk_host, making it easier to see what crates compile ziskos in host mode or guest mode

### DIFF
--- a/examples/sha-hasher/guest/Cargo.toml
+++ b/examples/sha-hasher/guest/Cargo.toml
@@ -7,4 +7,4 @@ edition = "2024"
 byteorder = "1.5.0"
 sha2 = "0.10.8"
 serde = { version = "1.0", default-features = false, features = ["derive"] }
-ziskos = { path = "../../../ziskos/entrypoint" }
+ziskos = { path = "../../../ziskos/entrypoint", default-features = false, features = ["zisk_guest"] }
--- a/ziskos/entrypoint/src/syscalls/arith256_mod.rs
+++ b/ziskos/entrypoint/src/syscalls/arith256_mod.rs
@@ -1,9 +1,9 @@
 //! Arith256Mod system call interception
 
-#[cfg(all(target_os = "zkvm", target_vendor = "zisk"))]
+#[cfg(feature = "zisk_guest")]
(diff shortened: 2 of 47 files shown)


examples/sha-hasher/guest/Cargo.toml
```diff
@@ -7,4 +7,4 @@ edition = "2024"
 byteorder = "1.5.0"
 sha2 = "0.10.8"
 serde = { version = "1.0", default-features = false, features = ["derive"] }
-ziskos = { path = "../../../ziskos/entrypoint" }
+ziskos = { path = "../../../ziskos/entrypoint", default-features = false, features = ["zisk_guest"] }
```
kevaundray, line 10: This is the main change for guest programs -- we could make zisk_guest be default and then host compilations would need to pass default-features=false

ziskos/entrypoint/src/syscalls/arith256_mod.rs
```diff
@@ -40,9 +40,9 @@ pub extern "C" fn syscall_arith256_mod(
     params: &mut SyscallArith256ModParams,
     #[cfg(feature = "hints")] hints: &mut Vec<u64>,
 ) {
-    #[cfg(all(target_os = "zkvm", target_vendor = "zisk"))]
+    #[cfg(feature = "zisk_guest")]
```
kevaundray, line 43: Most changes look like this -- note that for supporting rv64imac-unknown-none-elf, these flags would not change

hecmas:
Superseded by PR #1007.

PR #803, opened 2026-02-18
chore(no_std): Add allocator for `target_os = none`

kevaundray:

In order to support rv64imac-unknown-none-elf or rv64im-unknown-none-elf in the future, the code would require an implementation of a global allocator. This re-uses the bump allocator that is already being used for `sbrk`


hecmas:
Superseded by #965.

PR #814, opened 2026-02-24
Double input max size

fractasy:


fractasy:
Closed as we already support 1GB input size since v0.16.0

PR #834, opened 2026-03-05
increase inputdata to 1gb

zkronos73:


zkronos73:
changes merged on pre-develop-0.16.0

PR #837, opened 2026-03-06
Feature/input stream

fractasy:


fractasy:
Already merged as part of the release 0.16.0 process, so closing

PR #852, opened 2026-03-12
fix: adding bls range checks

hecmas:
This pull request strengthens input validation for BLS12-381 curve operations by ensuring all point coordinates and field elements are checked to be within the field before curve membership or subgroup checks. It also updates result codes to reflect new error conditions for invalid field elements and propagates these checks across G1, G2, MSM, map-to-curve, and pairing operations.

These changes ensure that all curve-related operations fail early and clearly when given invalid field elements, improving safety and correctness.


RogerTaule (approved):

xavi-pinsach (commented):
LGTM

PR #856, opened 2026-03-13
fix bug with reentrant-recursive edge case and develop call stack disable on error

zkronos73:
Fix edge case of callstack simulation on recursive calls. (stats)


hecmas (approved):
LGTM

PR #864, opened 2026-03-17
Remove unwanted trace

fractasy:


fractasy:
Close since the same bug has already been fixed in pre-develop-16.1

PR #870, opened 2026-03-19
RFC: `user_api_reference.md` document

xavi-pinsach:


book/developer/use_api_reference.md
```diff
@@ -0,0 +1,400 @@
...
+    metadata:    Option<String>, // JSON
+}
+
+struct RegisterGuestProgramResponse {
+    program_id: String, // UUID
+    hash_id:    String, // hash of zisk_elf
+    status:     ProgramStatus,
+}
+```
+
```
xavi-pinsach, line 141: > [!NOTE]
> I propose introducing a new call, WaitGuestProgram, to poll a guest program for readiness.
>Its behavior would be similar to WaitJobResult, providing a blocking mechanism with a timeout to wait until the program becomes ready.

### `WaitGuestProgram`

The intended primitive for polling a program to readiness. The server holds the response for
up to 5 seconds: if the program reaches a terminal state (`Ready` or `Failed`) within that
window, it returns immediately; if the 5 seconds elapse first, it returns with the current
status and the client can re-issue.

```
WaitGuestProgramRequest → GuestProgramSummary
```

```rust
struct WaitGuestProgramRequest {
    program_id: String,
}
```

book/developer/use_api_reference.md
```diff
@@ -0,0 +1,400 @@
...
+```rust
+struct DeleteGuestProgramRequest {
+    // exactly one of program_id or hash_id must be supplied
+    program_id: Option<String>,
+    hash_id:    Option<String>,
+}
+```
+
+## Proof Requests
```
jestape, line 185: We'll have to add a execute request for tracing in a operating environment how the hints / input stream is being consumed. This could either (A) be a new ExecuteRequest or (B) a ProofKind. The call should return a trace of consumption or a at least a counter of time where the process was stalled in the finputcall.

xavi-pinsach: On your tracing suggestion I'd like to propose an alternative API shape that covers this and more.
> [!IMPORTANT]
> **The core idea: split dispatch from observation.**

Instead of a blocking ExecuteRequest that returns a stream, I'd propose:
  1. `ProveRequest` — fires immediately, returns a job_id. No waiting.
  2. `WatchJob(job_id, trace_level)` — a server-streaming RPC the client subscribes to after dispatch. Server pushes JobEvent messages for the lifetime of the job.

  Why this is better than a blocking ExecuteRequest or a new ProofKind:

  - Decoupled concerns: The request describes what to run. The watch describes what to observe. Tracing granularity is a client/consumer concern — it shouldn't influence how the job
   is dispatched server-side.
  - Multiple observers watches on the same job, simultaneously, without coordination.
  - Reconnect resilience: If the client connection drops mid-job, it re-attaches with WatchJob. A single blocking RPC loses everything on disconnect.

  Regarding your example: The stall counters from finputcall fit naturally as events in the watch stream, gated by trace_level. The worker already knows when it's blocked on input and just needs to emit that as an event.

PR #872, opened 2026-03-19
feat: Accelerator spec conformance

marcinbugaj:
The other PR is https://github.com/0xPolygonHermez/zisk-eth-client/pull/53

This PR brings ziskos into conformance with the zkVM accelerator C-interface standard and adds a new crate for consuming the interface from C.                                                                                
                                                                                                                                                                                                                                
  Changes:                                                                                                                                                                                                                      
                                                                                                                                                                                                                              
  - ripemd160 accelerator (ziskos/entrypoint): Added zkvm_ripemd160 — a pure-software RIPEMD-160 implementation exported via the standard C interface, completing the set of EVM precompile accelerators.                       
  - Standard C interface (zkvm_accelerators.h + zkvm_accelerators.rs): Introduced the full zkvm_accelerators.h header and matching Rust implementations for all 19 standard accelerator functions. All function signatures, type
   names, and return codes now exactly match the standard. A bindgen-based build script generates type bindings from the header at compile time, providing compile-time verification that the Rust implementations match the C  
  ABI.                                                                                                                                                                                                   
  - ziskos-static-lib (new crate): A staticlib crate that exports zkvm_* functions as a C static library (.a). Built with fat LTO and single codegen unit for maximum dead-code elimination. Intended for linking   
  ziskos accelerators into C/C++/C#/Go guest programs targeting the ZisK zkVM.


Cargo.toml
```diff
@@ -60,6 +61,10 @@ opt-level = 3
 [profile.release.package."proofman-verifier"]
 opt-level = 0
 
+[profile.release.package.ziskos-static-lib]
+opt-level = 3
```
hecmas: This is redundant as it is the workspace default.

ziskos-static-lib/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+
+[dependencies]
+ziskos = { workspace = true }
+
+# When building this crate as a standalone package (outside the workspace),
+# use fat LTO and a single codegen unit for maximum dead-code elimination.
+[profile.release]
+opt-level = 3
+lto = "fat"
+codegen-units = 1
```
hecmas: You should specify the profile at the workspace root, otherwise it is ignored by the compiler:
```bash
$ cargo build
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /home/hector/zisk/zisk2/ziskos-static-lib/Cargo.toml
workspace: /home/hector/zisk/zisk2/Cargo.toml
```

ziskos/entrypoint/src/zisklib/lib/ripemd160.rs
```diff
@@ -0,0 +1,24 @@
...
+    output
+}
+
+/// C-compatible wrapper for full RIPEMD160 hash
+///
+/// # Safety
+/// - `input` must point to at least `input_len` bytes
+/// - `output` must point to a writable buffer of at least 32 bytes
+#[inline]
+pub(crate) unsafe fn ripemd160_c(input: *const u8, input_len: usize, output: *mut u8) {
```
hecmas: Similar to the rest of precompiles, we should (feature-based) add hints to it.

ziskos/entrypoint/src/zisklib/lib/bn254/mod.rs
```diff
@@ -21,3 +21,8 @@ pub use fp6::*;
 pub use fr::*;
 pub use pairing::*;
 pub use twist::*;
+// Success/failure codes exposed to zkvm_accelerators (pub(crate) to avoid glob-re-export conflicts with bls12_381)
+pub(crate) use curve::{
+    G1_ADD_SUCCESS, G1_ADD_SUCCESS_INFINITY, G1_MUL_SUCCESS, G1_MUL_SUCCESS_INFINITY,
+};
+pub(crate) use pairing::{PAIRING_CHECK_FAILED, PAIRING_CHECK_SUCCESS};
```
hecmas: I think these re-exports are redundant as we export the entire files. Maybe we should uniform all these exports in a future.

ziskos/entrypoint/src/zisklib/lib/secp256k1/ecdsa.rs
hecmas, line 1: We should try to clean the (now deprecated) `secp256k1_ecdsa_verify_and_address_recover_c` and `secp256k1_ecdsa_address_recover_c` functions.

hecmas (changes requested):

Cargo.toml
marcinbugaj: Done

ziskos-static-lib/Cargo.toml
marcinbugaj: Done

ziskos/entrypoint/src/zisklib/lib/secp256k1/ecdsa.rs
marcinbugaj: Done

ziskos/entrypoint/src/zisklib/lib/ripemd160.rs
marcinbugaj: Done

ziskos/entrypoint/src/zisklib/lib/bn254/mod.rs
marcinbugaj: Done

hecmas (approved):
LGTM

agnusmor (approved):

xavi-pinsach (changes requested):
A couple of architectural concerns before we merge:

#### 1. `ziskos-hints/build.rs` — cross-crate file reference

```rust
let header = format!("{}/../ziskos/entrypoint/zkvm_accelerators.h", manifest_dir);
```

This relative path reference to a sibling crate is fragile and it will break crates.io publishing. At the same time there is duplicated logic in both `build.rs`.

The bindings could be generated once in ziskos and re-exported or accessed from ziskos-hints?

#### 2. ziskos-static-lib

Can we avoid the indirection and keep the workspace cleaner?

The entire crate is:

```rust
pub use ziskos::zisklib::lib::zkvm_accelerators::*;
```

can be avoided by adding staticlib to ziskos's crate type?

```rust
# ziskos/entrypoint/Cargo.toml
[lib]
crate-type = ["rlib", "staticlib"]
```

marcinbugaj:
> A couple of architectural concerns before we merge:
> 
> #### 1. `ziskos-hints/build.rs` — cross-crate file reference
> 
> #### 2. ziskos-static-lib
>

@xavi-pinsach, these issues were addressed:

Ad 1. A separate package "zkvm-interface" was created that both "ziskos" and "ziskos-hints" depend on. Commit: [8b10062](https://github.com/0xPolygonHermez/zisk/pull/872/commits/8b1006281cc06ce2005cf1c47d32fe3b354ecedc)

Ad 2. The suggested approach works and it was applied: [c974344](https://github.com/0xPolygonHermez/zisk/pull/872/commits/c974344f2f61636609356385b2eb77b151c64712). The only disadvantage of it is that the control over the exports is not that fine grained which results in bigger libraries. For example, 
```
cargo +zisk build -p [ziskos-static-lib or ziskos] --release --target riscv64ima-zisk-zkvm-elf --config 'profile.release.lto="fat"'
```

gives:
```
39M  Mar 24 16:28 libziskos.a
2.9M Mar 24 16:28 libziskos_static_lib.a
```

PR #889, opened 2026-04-07
chore: use global buffer for syscall hints

kevaundray:
Removes the threading of hints across many functions.

It seems that the word hints is used differently in two places:

- precompile input: The first usage logs the inputs to precompiles, this is logged into `HINT_BUFFER` and eventually saved to .bin file

- syscall output: The second is what this PR edits, its the output for the syscalls that are made. This is not saved to disk.

Example: If we were doing `sha256(a,b,c)`, we would save `a,b,c` to disk since its the precompile input. We would then save the output of the syscall `sha256f`. A single sha256 call can have many sha256f calls so we would actually save multiple permutation calls.

It may make sense to rename the usages of hint, if the above understanding is correct

### DIFF
--- a/ziskos/entrypoint/src/hints_collect.rs
+++ b/ziskos/entrypoint/src/hints_collect.rs
@@ -1,0 +1,46 @@
+#[cfg(feature = "hints")]
+use std::cell::{Cell, RefCell};
+
+#[cfg(feature = "hints")]
+thread_local! {
+    static BUFFER: RefCell<Vec<u64>> = RefCell::new(Vec::with_capacity(64));
+    static DEPTH: Cell<u32> = Cell::new(0);
+}
+
+/// Append a single u64 to the hints buffer.
+#[cfg(feature = "hints")]
+#[inline(always)]
+pub fn hints_push(val: u64) {
+    BUFFER.with(|b| b.borrow_mut().push(val));
+}
+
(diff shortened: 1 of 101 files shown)


ziskos/entrypoint/src/hints_collect.rs
```diff
@@ -0,0 +1,46 @@
+#[cfg(feature = "hints")]
+use std::cell::{Cell, RefCell};
+
+#[cfg(feature = "hints")]
+thread_local! {
+    static BUFFER: RefCell<Vec<u64>> = RefCell::new(Vec::with_capacity(64));
```
kevaundray, line 6: Adds a global buffer for syscall outputs. Note that we have something like this already for precompile inputs named `HINT_BUFFER`

hecmas:
> It may make sense to rename the usages of hint, if the above understanding is correct

I personally use "hint" as a generic concept in the host-guest relationship: a hint is anything that lives in the host memory (independent but not disjoint from the guest memory) and that is sent from the host to the guest on request to speed things up. Probably we should define some suffixes for the different hint flavors. For example, there is a third use of hints that you do not list and is what we refer in the code as *free-inputs* (use to "verify" instead of "compute") and have a different purpose than the execution hints.

hecmas (approved):
LGTM

agnusmor (approved):

PR #907, opened 2026-04-17
Add utils and clean-cache commands and refactor clean command

agnusmor:


cli/src/commands/check_setup.rs
```diff
@@ -22,9 +22,9 @@ pub struct ZiskCheckSetup {
     #[arg(short = 'w', long)]
     pub proving_key_plonk: Option<PathBuf>,
 
-    /// Enable proofs aggregation
+    /// Disable proofs aggregation
```
xavi-pinsach, line 25: Add to the original aggregation flag a default value to true could be better to keep backwards compatibility

cli/src/commands/utils.rs
```diff
@@ -0,0 +1,135 @@
...
+    pub fn run(&mut self) -> Result<()> {
+        setup_logger(VerboseMode::Info);
+
+        print_banner();
+        print_banner_command("Clean");
+
+        let home_zisk_path = get_home_zisk_path()?;
+        let cache_zisk_path = home_zisk_path.join("cache");
+
+        if cache_zisk_path.exists() {
```
xavi-pinsach, line 60: if !cache_zisk_path an Err could to be returned insted of Ok(())

PR #914, opened 2026-04-22
Fix: update function name in `zkvm_accelerators.rs`

marcinbugaj:
Commit d80a5cd62 ("Adding k256 functions") renamed secp256k1_ecdsa_verify_c → secp256k1_ecdsa_verify_bytes_c in secp256k1/ecdsa.rs and updated one call site in zkvm_accelerators.rs, but missed a second call at line 1091 inside a #[cfg(all(target_os = "zkvm", target_vendor = "zisk"))] block. That block is excluded from host builds, so the bug was invisible until cargo-zisk compiled for the zkvm target.


hecmas (approved):
Thanks

PR #917, opened 2026-04-30
Adding zkvm_init and zkvm_deinit

RogerTaule:
This PR contains 5 changes:

- Removes read_input for the zkvm + zisk target and fixes the corresponding import in io.rs
- Adds zkvm_init and zkvm_deinit so these functions can be used in C guest programs
- Wires these functions into the entrypoint! macro, so the user doesn't have to add them in the guest — they are inserted automatically when using ziskos::entrypoint!(main) and compiling with the zisk_hints flags
- Adds environment variables so we can specify the input file for native reads (instead of always using build/input.bin) and also specify the output file for hints
 - Adds a set_native_input call so that on native we can pass an in-memory region rather than only reading from a file

### DIFF
--- a/book/getting_started/hints_stream.md
+++ b/book/getting_started/hints_stream.md
@@ -327,36 +327,66 @@ As a result, for each `zisklib` function invocation, the `HintsProcessor` may ge
 
 ### 5.2 Initialize/Finalize Hint Stream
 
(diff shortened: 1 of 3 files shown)


hecmas (approved):
LGTM

agnusmor (approved):

PR #935, opened 2026-05-05
Fix overlapping data sections in elf_extraction.rs

fractasy:


fractasy:
Waiting for Kev to review

kevaundray:
Ah I can see why the test fails — I think its maybe subjective as to what we do.

Essentially, if we have two sections that overlap, we would want to define what the behaviour is:

- The extractor could error, given this should not happen
- The extractor could truncate one of the sections

An example:

- Chunk A: starts at address `0x1000`, contains 8 bytes: `[A0, A1, A2, A3, A4, A5, A6, A7]` (covers 0x1000–0x1007)
- Chunk B: starts at address `0x1002`, contains 2 bytes: `[B0, B1]` (covers 0x1002–0x1003)

With this PR, it seems that the result in ROM would be `[A0, A1, B0, B1]` so `A4` to `A7` gets truncated.

A different strategy is to only overlap on the parts that conflict, so it would become `A0, A1, B0, B1, A4, A5, A6, A7`

Before this PR, we were leaving it upto whomever consumed the ROM sections (mainly to not change the previous behaviour), but the right behaviour here may be to just error out if it happens, since I think it shouldn't happen on a well behaved linker

fractasy:
After discussing the issue internally, we don't support overlapped sections, and we think it's the simplest, safer approach.

fractasy:
Closing without merging.

PR #939, opened 2026-05-05
feat: Implement zkVM standards I/O interface

rubo:
_I presume some changes are debatable, so feel free to change them however you see fit. The goal is to have it supported in ZisK in one way or another._

## Summary

Implements the [zkVM Standards IO interface](https://github.com/eth-act/zkvm-standards/tree/main/standards/io-interface) for ZisK:

- Adds `zkvm_io.h` with:
  - `read_input(const uint8_t** buf_ptr, size_t* buf_size)`
  - `write_output(const uint8_t* output, size_t size)`
- Extends `zkvm-interface` bindgen generation to include the IO header.
- Adds ZisK implementations in `zisklib::zkvm_io`.
- Routes `ziskos::io::commit_slice` through the new standard output writer so both APIs share one output cursor.
- Adds a compile-time ABI check that statically asserts the Rust implementations match the bindgen-generated signatures.

### Input

ZisK stdin is length-prefixed and record-based, while the standard expects one idempotent private input buffer.

This PR exposes the first logical ZisK stdin record as the standard input buffer:

- zkVM target: zero-copy pointer into `INPUT_ADDR`.
- native target: reads once, caches, and returns a stable pointer.
- repeated `read_input` calls return the same buffer.
- standard reads do not advance ZisK's streaming input cursor.

The Rust and C docs explicitly warn not to mix standard IO reads with ZisK streaming reads for the same input.

### Output

`write_output` appends byte buffers as one continuous stream:

- full `u32` public-output slots are written directly;
- partial slots are buffered and coalesced across calls;
- zero-length writes are no-ops.

`commit_slice` now delegates to this same writer, removing duplicate output cursor state.

## Reservations / Behavior changes

- `commit_slice` no longer pads each call independently. Partial writes across multiple calls now coalesce. This matches the standard and is more efficient, but it is a behavior change.
- Native `read_input` intentionally leaks cached input buffers on reset to keep returned raw pointers valid. This is native/test-only.


RogerTaule:
@cla-bot check

marcinbugaj:
I have concern about the (in)compatibility of the following input access functions:
- `extern "C" read_input` from the standard AND
- `zisk::io::read`

The latter is already used [here](https://github.com/0xPolygonHermez/zisk-eth-client/blob/0716e42eb7795ee85b466dc7778660052e16019d/bin/guests/stateless-validator-reth/src/main.rs#L20) and [here](https://github.com/0xPolygonHermez/zisk-eth-client/blob/0716e42eb7795ee85b466dc7778660052e16019d/bin/guests/stateless-validator-reth/src/main.rs#L41) in the reth guest program in `zisk-eth-client`.

The files in [here](https://github.com/0xPolygonHermez/zisk-eth-client/tree/main/bin/guests/stateless-validator-reth/inputs) have the following binary format:
```
[8B len1][pub bytes][pad][8B len2][wit bytes][pad]
```

Sadly, those files cannot be processed using the  `extern "C" read_input` from the standard. The reason is that the function can access only the first "packet": `[8B len1][pub bytes]`. The other bytes from the file are inaccessible to `extern "C" read_input`. One solution is to modify [input gen](https://github.com/0xPolygonHermez/zisk-eth-client/tree/main/bin/input-gen) to prepend the length of the buffer as `outer_len`:

```
[8B outer_len][8B len1][pub bytes][pad][8B len2][wit bytes][pad]
```

That change would make the file incompatible with `zisk::io::read` though (forcing manual deserialization of the inner packets)

What's your take on this issue?

My expectation for `extern "C" read_input` when running with a file on disk would be that it returns a pointer to a buffer containing the full file contents, with the reported size equal to the actual file size - rather than reading the size from the first 8 bytes of the file itself.

rubo:
> I have concern about the (in)compatibility

I share your concerns. That's why the PR description starts with "some changes are debatable". @RogerTaule also wants to discuss the standards with EF. Thus, we need a broader discussion to improve the standard, as it limits Zisk functionality. On the other hand, different zkVMs may have very different features, and no standard can cover them all. So, we need some good middle ground.

> What's your take on this issue?

Speaking of standards, my take is to follow the [input format](https://github.com/ethereum/execution-specs/blob/projects/zkevm-bal-devnet-3/src/ethereum/forks/amsterdam/stateless_ssz.py#L158-L164) proposed by the standard, so you don't need to specify the witness separately and read it in another go (if I got it right). With that, it works with a single read using the standard I/O interface. I'm not saying it's better or worse, but it's the standard at the moment.

PR #941, opened 2026-05-06
Fix: 4 constraint fixes for non-Rust guest programs (unaligned memory, ELF sections, bus accounting)

juanfranblanco:
## Summary

  Five fixes for constraint violations that affect any non-Rust guest program targeting Zisk (C, C++, Go, .NET/NativeAOT, Zig). Rust programs avoid these because the compiler enforces natural alignment and
  Zisk's test suite is Rust-only.

 Issue / conversation: https://github.com/0xPolygonHermez/zisk/issues/931

  Discovered while proving Nethereum/Ethereum EVM execution (C# via NativeAOT → RISC-V → Zisk). All fixes verified with `cargo-zisk verify-constraints -d` and full prove+verify cycles across 4 witness variants
  (3.96M–8.04M steps, 1–2 Main instances),  against 0.17.0.

  ## Fixes

  ### 1. `get_read_value` double shift — `state-machines/mem-common/src/mem_helpers.rs`
  Cross-word-boundary reads applied `>> offset` twice: `read_values[1] << (64 - offset) >> offset` → `read_values[1] << (64 - offset)`. Caused ~104 MemAlign constraint violations.

  ### 2. Read-before-write in MemAlign TwoWrites — `state-machines/mem/src/mem_align_sm.rs`
  The TwoWrites path read `first_write_row.get_reg(i)` and `second_write_row.get_reg(i)` before those rows were populated. Fixed by computing values directly from `value_first_write`/`value_second_write` using
   `Self::get_byte()`.

  ### 3. `is_full_aligned` wrong mask — `core/src/mem.rs`
  Alignment check used `addr & 0x03` (4-byte mask) instead of `addr & 0x07` (8-byte mask) in 3 locations (`is_full_aligned`, `read_required`, `write_required`). An 8-byte read at `0xAFFF1B04` passes the 4-byte
   check but crosses a word boundary. Caused 18/160 `read_same_addr` Mem constraint violations.

  ### 4. Hardcoded `bytes=8` in cross-word-boundary sub-word bus payload — `emulator/src/emu.rs`
  **Root cause of the previously reported multi-instance global constraint failure.** In `source_b_mem_reads_consume_databus()`, the `SRC_IND` double_not_aligned paths (read line 847, write line 1281)
  hardcoded `8` in the bus payload width instead of `instruction.ind_width`. Sub-word operations (LW/LH/LB at offsets where `offset + width > 8`) emitted `bytes=8` on the bus, but Main PIL assumes the actual
[...]

PR #962, opened 2026-05-10
Fix: NativeAOT alignment processing issues

maximmenshikov:
# Summary
While investigating operation of NativeAOT applications, like [Nethermind](https://github.com/nethermindeth/nethermind), we found several issues while proving the execution.

## Issue 1 (f9998a87f48ed825b6b703ea4730e7f89e7e33d1)

Patch fixes usage of invalid value, causing global constraints to fail.

## Issue 2 (5db4e39c4d1e15dca9bab063e5a1b0acce2fc0be)

Patch fixes spontaneous loss of the second word during unaligned 8-byte access.

More information in particular commits.

The correct operation of NativeAOT also needs @juanfranblanco patches from the PR #941. For ease of check, I suggest to verify branch [feature/nethermind_test](https://github.com/NethermindEth/zisk/tree/feature/nethermind_test) in our Zisk fork. It already contains cherry-picked patches from @juanfranblanco. The particular binary and the super small test block we used is attached.
[nethermind.zip](https://github.com/user-attachments/files/27566589/nethermind.zip)
[150026.bin.zip](https://github.com/user-attachments/files/27566599/150026.bin.zip)

## Results
These patches, along with other mentioned ones, allow full end-to-end proving for Nethermind NativeAOT client on both emulator and GPU.

## cla
Confirming that I have read cla and agree to that.

### DIFF
--- a/core/src/zisk_rom_2_asm.rs
+++ b/core/src/zisk_rom_2_asm.rs
@@ -1999,6 +1999,14 @@ impl ZiskRom2Asm {
                                 *code += &format!("pc_{:x}_b_address_check_done:\n", ctx.pc);
                             }
                             4 | 2 => {
+                                // Save original addr to AUX before &~7 destroys it
+                                *code += &format!(
+                                    "\tmov {}, {} {}\n",
+                                    REG_AUX,
+                                    reg_address,
+                                    ctx.comment_str("aux = original address")
+                                );
+
                                 // Calculate previous aligned address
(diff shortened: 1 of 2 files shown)


zkronos73:
@cla-bot check

zkronos73 (approved):

fractasy (approved):
The bug was in the calculation of the next, upper aligned address.  God catch.

PR #804, opened 2026-02-18
chore(no_std): Make `ziskos` crate `no_std` compatible

kevaundray:
This pulls in the changes from:

- #800
- #801
- #802
- #803

And adds commit 6446d86 to replace all places that import std and gate the std/host only dependencies

The above PRs could be merged in indepdently, they seemed easier to pull out of everything else and review.

PR #800, opened 2026-02-18
chore(no_std): update bincode to v2.0

kevaundray:
Related to https://github.com/0xPolygonHermez/zisk/issues/793

In order to support rv64imac-unknown-none-elf or rv64im-unknown-none-elf in the future, ZiskOS would need to be `no_std` compatible -- upgrading bincode v2 is a step towards this.

PR #965, opened 2026-05-11
feat: `ziskos` made `no_std` compatible for ZisK target

marcinbugaj:
### Motivation

`libziskos.a` is a C static archive. As a `staticlib` crate, the Rust toolchain bundles all transitive object files into it — including those from any dependency that pulls in `std`. Before this change, `zkvm-interface` caused 16+ std CGU object files to be included, because its bindgen-generated bindings referenced `::std::os::raw::c_int`.

### What changed

**`ziskos` is now `no_std` for the zkVM target.**

- Added `#![cfg_attr(zkvm, no_std)]` to `lib.rs`.
- Restructured `Cargo.toml` to split dependencies by target. `fields`, `ark-*`, `blst`, `secp256k1`, and other heavy host-only crates are now excluded from the zkVM build entirely.
- `fcalls_impl` (software fallback implementations that depend on those crates) is gated to non-zkvm.
- `sys_rand` replaced `lazy_static` + `std::sync::Mutex<StdRng>` with `spin::Mutex<Option<SmallRng>>` + `AtomicBool` - both `no_std` compatible.
- Fixed stray `use std::` imports across ~18 files (`core::arch::asm`, `core::ptr`, `core::mem::MaybeUninit`, etc.) and added `use crate::alloc_extern::vec::Vec` where needed. The `alloc` crate is linked as `extern crate alloc as alloc_crate` to avoid colliding with the existing `mod alloc` bump-allocator module.
- **`zkvm-interface`**: Adding `.use_core()` to the bindgen builder switches the generated C types from `::std::os::raw` to `::core::ffi`, after which the crate compiles without `std` and `#![no_std]` can be added.
- **`zisk-definitions`**: Contains only `const` declarations with no std usage; `#![no_std]` is added trivially.
- **`ziskos` — panic handler**: Making `zkvm-interface` `no_std` removed the last dependency that implicitly linked the Zisk custom `std`, which had been providing the `#[panic_handler]` lang item that the compiler requires for every `no_std` static library. A panic handler is added to `ziskos` inside the existing zkvm-gated module. It writes the panic message to UART then calls `core::intrinsics::abort()`, which the Zisk LLVM backend lowers to the RISC-V `unimp` instruction — matching the `abort_internal()` implementation in the Zisk standard library. This handler serves both consumers of `ziskos`: Rust crate dependents (via the rlib) and C programs linking the archive (where `rust_begin_unwind` must be resolvable from within the archive itself).

**Bincode v1 → v2 (workspace-wide).**

Bincode v1 has no `no_std` support, so enabling `io::read`/`commit` on zkVM required upgrading. Rather than carrying two majors, the entire workspace was migrated to v2. All call sites in `common`, `sdk`, and `distributed` updated to the new API. The `io.rs` cfg-split is gone — both targets now share a single code path.

> **Note**: [Bincode 1 and 2 are completely compatible if the same configuration is used](https://crates.io/crates/bincode/2.0.0)

### CI check

A step is added to the `test-x86_64` job that builds `ziskos` for the zkvm target and inspects `libziskos.a` with `ar t`. Since Rust embeds the crate name in codegen unit filenames, any bundled `std` object file is immediately identifiable by name. The check fails the build and prints the offending filenames if any are found.


marcinbugaj:
This PR supersedes:
- https://github.com/0xPolygonHermez/zisk/pull/800
- https://github.com/0xPolygonHermez/zisk/pull/801
- https://github.com/0xPolygonHermez/zisk/pull/803
- https://github.com/0xPolygonHermez/zisk/pull/804

ziskos/entrypoint/src/zisklib/lib/bn254/curve.rs
```diff
@@ -1,5 +1,9 @@
 //! Operations on the BN254 curve E: y² = x³ + 3
 
+#[cfg(all(target_os = "zkvm", target_vendor = "zisk"))]
+use crate::alloc_extern::vec::Vec;
+
+#[cfg(not(all(target_os = "zkvm", target_vendor = "zisk")))]
 use num_traits::ops::bytes;
```
hecmas: We can remove this. Still there from an older version.

ziskos/entrypoint/Cargo.toml
```diff
@@ -14,40 +14,41 @@ crate-type = ["rlib", "staticlib"]
 zkvm-interface = { workspace = true }
 lib-c = { workspace = true }
 
+getrandom = { version = "0.2", features = ["custom"] }
+cfg-if = "1.0"
+ripemd = { version = "0.1.3", default-features = false }
```
hecmas, line 19: I assume `ripemd` is `no_std`.

hecmas (commented):
Just a couple of comments. LGTM overall.

RogerTaule (approved):
LGTM too! Btw, since bincode is no longer mantained, which alternatives would you propose? Thanks for the PR :)

ziskos/entrypoint/Cargo.toml
marcinbugaj: It seems so: https://docs.rs/crate/ripemd/0.2.0/source/src/lib.rs
There's "#![no_std]" at the beginning of lib.rs :)

ziskos/entrypoint/src/zisklib/lib/bn254/curve.rs
marcinbugaj: Could I do that in subsequent PRs? If I push a change now then it would take ~40 minuts of CI job more :) Otherwise it could be merged now if you accept

ziskos/entrypoint/src/lib.rs
```diff
@@ -452,4 +459,13 @@ pub mod ziskos {
...
     //core::arch::global_asm!(include_str!("dma/inputcpy.s"));
     core::arch::global_asm!(include_str!("dma/memset.s"));
+
+    #[panic_handler]
+    fn panic(info: &core::panic::PanicInfo) -> ! {
+        if let Some(msg) = info.message().as_str() {
+            sys_write(1, msg.as_ptr(), msg.len());
+            sys_write(1, b"\n".as_ptr(), 1);
+        }
+        core::intrinsics::abort()
```
hecmas, line 470: Attempting to compile a guest that uses the std produces a duplication error since the `panic` implementation is imported by default by the std. We should either remove or gate this function through a feature.

hecmas (approved):
LGTM

PR #971, opened 2026-05-12
Fix bug in final lt_flag verification in ArithEq and ArithEq384. Optimization to save one witness column

zkronos73:
> Thanks to @jsign for raising this up.

Fixed a bug in the PIL code for ArithEq and ArithEq384 where the constraint to verify that x3_lt_flag and y3_lt_flag equal 1 at the end of each cycle was missing.

To fix this issue and also optimize the implementation, the transition function verification table for lt_flag has been split into three parts:

- **Middle clocks**: The table that existed previously, with standard borrow propagation.
- **First clock**: Similar to the middle clocks table, but ignoring the 'lt_flag value when computing lt_flag.
- **Last clock**: Contains only transitions to lt_flag = 1, as any other value would be invalid.

The last clock split ensures that lt_flag equals 1 at the end of the cycle. Without this constraint, an invalid final state wouldn't match the lookup and proof generation would fail, effectively enforcing the missing constraint.

The first clock split was introduced for optimization purposes, reducing the lookup degree from 2 to 1. This was achieved by eliminating the 'lt_flag * (1 - CLK_0) product and adding 2 intermediate witness columns for delta_x3 and delta_y3. Since reducing the degree saved 3 columns and adding the new columns cost 2, there's a net savings of 1 witness column.

Eliminating 'lt_flag * (1 - CLK_0) means we need to know the 'lt_flag value from the previous input when calculating multiplicity. Since inputs are processed in parallel, we now propagate the expected lt_flag (which depends solely on the operation type, not on the operand values) when sequentially generating input information.


hecmas (approved):
LGTM

PR #984, opened 2026-05-14
libziskos_standalone.a

marcinbugaj:
Separates the `libziskos.a` static library into a dedicated `ziskos-staticlib` workspace crate, and adds a CI check to verify the archive is correct. I think that a separate crate is even better because it allows fine grained export of functions/symbols and the resulting archive is smaller.

### Changes

- `ziskos/entrypoint` is now `crate-type = ["rlib"]` only, with no `#[panic_handler]`.
- A new `ziskos-staticlib` crate (alongside `ziskos-hints` at the workspace root) is `crate-type = ["staticlib"]`. It:
  - Explicitly re-exports the public interface of `ziskos` (`read_input`, `write_output`, `zkvm_init`, `zkvm_deinit`, all `zkvm_*` precompiles) to guarantee those symbols are bundled into the archive.
  - Provides the `#[panic_handler]` behind a `panic-handler` feature flag (enabled by default, disableable for consumers that supply their own).

### CI check

A new script `.github/scripts/check-libziskos.sh` is added and called from `pr.yml`. It builds `ziskos-staticlib` with `lto="fat"`, then verifies:

- No `std` object files are bundled in the archive.
- All required symbols are present: `_start`, `read_input`, `write_output`, `zkvm_init`, `zkvm_deinit`, and all 18 `zkvm_*` precompile accelerators.

### DIFF
--- a/.github/scripts/check-libziskos.sh
+++ b/.github/scripts/check-libziskos.sh
@@ -1,0 +1,72 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+. "$HOME/.cargo/env"
+
+FAIL=0
+TARGET=riscv64ima-zisk-zkvm-elf
+
+cargo +zisk build -p ziskos-staticlib --release \
+  --target "$TARGET" \
+  --config 'profile.release.lto="fat"'
+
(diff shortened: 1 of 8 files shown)


RogerTaule (approved):
LGTM

PR #1007, opened 2026-05-18
`zisk_guest` cfg to control conditional compilation

marcinbugaj:
> Supersedes PR #802.

Conditional compilation for guest-only code was scattered across ~75 files using the verbose predicate `all(target_os = "zkvm", target_vendor = "zisk")`. This hardcodes a single target triple, making it impossible to add a second guest target (e.g. `riscv64im-unknown-none-elf` for bare-metal testing) without touching every call site.

### What this PR does

Introduces a single `zisk_guest` rustc cfg that abstracts over all supported guest targets. A `build.rs` is added to `ziskos/entrypoint`, `ziskos-staticlib`, and `ziskos-hints`; each emits `cargo:rustc-cfg=zisk_guest` when the build target matches:

```
(target_os == "zkvm" && target_vendor == "zisk")
|| (target_arch == "riscv64" && target_os == "none")
```

All source files then replace `all(target_os = "zkvm", target_vendor = "zisk")` with the shorter `zisk_guest`.

**Note:** `Cargo.toml` dependency sections cannot use build-script cfgs (Cargo limitation), so they continue to use explicit target-triple predicates. The `any(zkvm, riscv64-none)` form is used there to cover both guest targets. As a related fix, `lib-c` (a host-only, non-`no_std` crate) is moved from the unconditional `[dependencies]` table into the host-only target section so it is no longer pulled into guest builds.

### Changes

- **`build.rs`** (×3 — `ziskos/entrypoint`, `ziskos-staticlib`, `ziskos-hints`): auto-detect guest target and emit `zisk_guest` cfg; declare it via `rustc-check-cfg` to suppress "unexpected cfg" warnings
- **`Cargo.toml`** (`ziskos/entrypoint`): move `lib-c` to host-only target dep section; expand target predicates to `any(zkvm, riscv64-none)` for shared guest deps; remove now-redundant `cfg(target_vendor, values("zisk"))` from `[lints]`
- **`Cargo.toml`** (`ziskos-staticlib`, `ziskos-hints`): remove `cfg(target_vendor, values("zisk"))` lint suppression (now handled by `build.rs`)
- **Source files** (~70 files): mechanical replacement of `all(target_os = "zkvm", target_vendor = "zisk")` → `zisk_guest` (and the negated form)

### No behaviour change

This is a pure refactor for the existing `riscv64ima-zisk-zkvm-elf` target. The emitted code is identical; the cfg just has a shorter name and is set one level up in `build.rs`.


hecmas (approved):
Overall, LGTM.

The Cargo.toml limitation can be a scalability problem, but I see that the tendency is standarizing a single guest target and point to it rather than supporting multiple.

Thanks for the `lib-c` fix.

agnusmor (approved):

PR #1011, opened 2026-05-19
fix: fill only from ROM_ADDR

han0110:
Go's linker always reserves one 4 KB page at the start of the first `PT_LOAD` for `PT_PHDR + NOTE`, so for a Go guest, we have to use `-T 0x80001000` and accept `.text` one page above `ROM_ADDR`. The resulting `min_program_pc = ROM_ADDR + 0x1000` makes the branch table gap-fill loop to fill gap between end of BIOS and `ROM_ADDR + 0x1000`, that creates huge `.rodata` in the generated assembly.

This PR fixes it by adding `previous_key >= ROM_ADDR` to the predicate so the gap-fill only happens when both endpoints sit inside the user-code region.

### DIFF
--- a/core/src/zisk_rom_2_asm.rs
+++ b/core/src/zisk_rom_2_asm.rs
@@ -3691,7 +3691,9 @@ impl ZiskRom2Asm {
                 //   4N + 3
                 // 4(N+1)
                 //   ...
-                if (*key > ROM_ADDR) && (*key != (previous_key + 1) && (*key != FLOAT_LIB_ROM_ADDR))
+                if (previous_key >= ROM_ADDR)
+                    && (*key > ROM_ADDR)
+                    && (*key != (previous_key + 1) && (*key != FLOAT_LIB_ROM_ADDR))
                 {
                     for _ in previous_key + 1..*key {
                         *code += "\t.quad 0\n";


fractasy:
@han0110 , the address map must be complete in order to do dynamic jumps (jumps to addresses that are not known at transpilation time, i.e. they come from a computed value).  If the problem is that you get 4k of empty/zero mapped addresses, that is not an issue.  Removing it would break the dynamic jumps.  If the problem is that you get tons of mapped addresses, from the 0x1000 entry ones, to the ROM program ones, that's a different issue.  Can you explain what do you see in the generated assembly code?

han0110:
What I see in the asm for the Go elf ([sample](https://github.com/han0110/zisk/blob/tmp/go-elf/basic_go.elf)), `sorted_pc_list` contains `424776` entries spanning three regions (`max_bios_pc = 0x9edfc`, `min_program_pc = 0x80001000`, `max_program_pc = 0x87f171d4`). The last BIOS entry is `0x9edfc` and the first user entry is `0x80001000`.

The current predicate checks `*key > ROM_ADDR` but not `previous_key`, so on the BIOS to user transition (`previous_key = 0x9edfc`, `*key = 0x80001000`) it fires and emits `0x7ff62203 ≈ 2.15 billion` `.quad 0` lines (~17.2 GiB `.rodata`) between `map_pc_9edfc` and `map_pc_80001000`, but IIUC we jump to either low or high table dynamically, so the filling in between seems not necessary.

fractasy:
Then the fix makes perfect sense.  Approved.

fractasy (approved):

PR #1013, opened 2026-05-19
Enable `ziskos-staticlib` `riscv64im-unknown-none-elf` compilation

marcinbugaj:
These two commits prepare `ziskos` for a second guest target (`riscv64im-unknown-none-elf`) alongside the existing `riscv64ima-zisk-zkvm-elf`. The `riscv64im` target is bare-metal and lacks the RISC-V 'A' (atomic) extension, which made the previous `sys_rand` implementation — built on `spin::Mutex` and `AtomicBool` — fail to compile for it.

### Commit 1 — Get rid of `spin` crate dependency

`sys_rand` is called on a single-core guest; it has no concurrent callers at any point, so a mutex and atomic flag are unnecessary overhead. The implementation is replaced with plain `static mut`:

```rust
static mut RNG: Option<SmallRng> = None;
static mut SYS_RAND_WARNING: bool = false;
```

This removes the `spin` crate from the dependency tree entirely and makes the code compile on any RISC-V guest target, regardless of whether the 'A' extension is present.

### Commit 2 — CI check for `riscv64im-unknown-none-elf`

Adds a new `build-riscv64im` job to the PR workflow that compiles `ziskos-staticlib` against the new target:

```
cargo +nightly build -p ziskos-staticlib --release \
  --target riscv64im-unknown-none-elf \
  -Z build-std=core,alloc \
  --config 'profile.release.lto="fat"'
```

**Toolchain setup:** `riscv64im-unknown-none-elf` is a Tier 3 target — there is no pre-built `std` to download, so `rustup target add` is not applicable. Instead, `-Z build-std` compiles `core` and `alloc` from source using the `rust-src` component. No external cross-compiler is needed; Rust's bundled `rust-lld` handles RISC-V linking. The toolchain setup reduces to a single `rustup` call:

```
rustup toolchain install nightly --component rust-src --no-self-update
```

The job runs on `ubuntu-latest` and has no dependencies on the GPU runner or any heavy native libraries.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -47,6 +47,24 @@ jobs:
           RUSTFLAGS: -Copt-level=3 -Cdebug-assertions -Coverflow-checks=y -Cdebuginfo=0 -C target-cpu=native
           RUST_BACKTRACE: 1
 
+  build-riscv64im:
+    name: Build ziskos-staticlib (riscv64im-unknown-none-elf)
+    runs-on: ubuntu-latest
+    env:
+      CARGO_NET_GIT_FETCH_WITH_CLI: "true"
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v4
+
+      - name: Install Rust nightly with rust-src
+        # riscv64im-unknown-none-elf is a Tier 3 target — no pre-built std is available.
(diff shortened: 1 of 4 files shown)


hecmas (approved):
LGTM

PR #1045, opened 2026-05-29
feat: lib-c macos compatibility

hecmas:


jbaylina:
This may look like a lot of changes to review. But mainly is the 
* ffiasm regenerated. 
* remove change linux specific types to generic and solve some naming conflicts in some headers. arith256, arith386 and add256
* lib-c Makefile

fractasy (approved):

PR #1060, opened 2026-06-02
docs: fix coordinator config key

mdqst:
the table was using `coordinator.port`, but the actual config examples use `coordinator.cluster_port`, so I updated it to match the real TOML key.


mdqst:
> We require contributors/corporates @mdqst to read our Contributor License Agreement, please check the [Individual CLA document](https://github.com/0xPolygonHermez/clabot-config/blob/master/CLA%20Zisk.pdf)

I have read and agree to the CLA

PR #1079, opened 2026-06-04
fix(rom-setup): key asm/ROM caches on their build inputs

RogerTaule:
The asm binaries and ROM/verkey were cached by guest-ELF hash alone, so a change to the transpiler, the embedded float ELF, or the linked C libs reused stale artifacts — including returning the wrong program vk.

Fold a fingerprint of those inputs into the cache filenames: build.rs bakes build-time hashes of the transpiler + float (ROM) and + emulator-asm sources (asm); asm_setup.rs adds the runtime lib contents; utils.rs keys the ROM/verkey on them. lib-c uses  so the hash is stable.

Adds unit tests, a before/after script, and a CI workflow.

### DIFF
--- a/.github/workflows/asm_cache_invalidation.yml
+++ b/.github/workflows/asm_cache_invalidation.yml
@@ -1,0 +1,59 @@
+name: asm cache invalidation
+
+# Guards that the rom-setup asm/ROM cache key tracks every input that determines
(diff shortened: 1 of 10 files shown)


xavi-pinsach (approved):
LGTM

PR #1091, opened 2026-06-08
fix(binary): soundness bug in LT_ABS_NP

hecmas:
> Thanks to @codygunton for [finding](https://github.com/codygunton/zisk/pull/5) the bug.

## Summary

`LT_ABS_NP` — used by the arithmetic SM to enforce `0 ≤ |remainder| < |divisor|` in signed division — returned `1` (less-than) instead of 0 when `|a| == |b|` and `|a|` is a multiple of 256. The two's-complement negation `|a| = ~a + 1` added the `+1` only at the least-significant byte and never propagated its carry, so at the magnitude-tie boundary the comparison used the wrong `|a|` bytes.

This is exploitable: a prover could submit a non-canonical division with remainder `r = -divisor` (e.g. `-256 / 256 → q=0, r=-256` instead of `q=-1, r=0`) whenever the divisor is a multiple of 256, bypassing the strict remainder bound and breaking quotient/remainder uniqueness.

## Fix

The negation carry is now propagated as its own chain. The per-byte carry packs two bits — `clt` (comparison) and `cneg` (negation) — the comparison runs on the true `|a|` byte (`(~a + cneg) & 0xFF`), and the negation overflow carries into the next byte. `LT_ABS_PN` is left unchanged (its `+1 `already lands in the subtrahend and is handled correctly); a comment now documents why.


zkronos73 (approved):
LGTM

PR #1097, opened 2026-06-10
fix: fill result with zeros when point addition or msm outputs infinity

han0110:
On `*_SUCCESS_INFINITY` the `zkvm_accelerators.rs` returns `ZKVM_EOK` without zero-filling the `result` (according to EIP-196 and EIP-2537, the point at infinity is encoded as all zero bytes). This PR fixes that so both successful paths have `result` written.


han0110:
Fixed by #1104

PR #1098, opened 2026-06-10
fix(pil): secure argument selectors

hecmas:
## Summary

The STD argument helpers (`lookup_assumes`, `permutation_assumes`, `permutation_proves`, `permutation`) constrain their selector to be binary **by default**, emitting `sel·(sel-1)===0`. In many of our state machines and precompiles the selector is already provably binary by construction (one-hot op selectors, latched flags, dedicated witness columns with their own `bits(1)` constraint), so that auto-added constraint is pure redundancy.

This PR threads the `sel_is_binary` opt-out flag through to those call sites and to the shared memory/operation/range helpers, so a verified selector can skip the duplicate constraint. It also fixes one latent soundness issue in `mem_align` that the opt-out would otherwise have turned into a live unconstrained-selector hole.

## Soundness fix — `mem_align` `sel_prove`

`sel_prove` is declared `col witness bits(1)`. The only constraints previously touching it were the disjointness `sel_prove * sel_assume === 0` and the value reconstruction. Its boundedness relied entirely on the STD `permutation`'s auto-added constraint on `sel_prove - sel_assume`.

That auto-constraint only forces `sel_prove - sel_assume ∈ {-1, 0, 1}`. On rows where `sel_assume = 0` (no aligned access) it therefore allowed `sel_prove ∈ {-1, 0, 1}` — i.e. `sel_prove = -1` was reachable, silently turning a "prove" row into an extra assume on the memory bus. `sel_prove` was never actually constrained to `{0, 1}`.

**Fix:** add the explicit `sel_prove * (sel_prove - 1) === 0`. This makes `sel_prove` a genuine binary selector; without it, opting out of the library constraint would leave `sel_prove` completely unconstrained on `sel_assume = 0` rows.

## Soundness notes

- Adding the opt-out **never removes a load-bearing constraint** at the touched sites: each selector is independently binary, verified per site.
- The helper default stays `0`, so opting out is always explicit and local.

## Misc
- `zisk_ops_costs.rs`: added a hash-throughput comparison comment (Poseidon2 / SHA2-256 / Blake2b / Keccak-256).
- `test-artifacts/programs/Cargo.lock`: dropped the unused `lib-float` entry.


zkronos73 (approved):
LGTM

PR #1107, opened 2026-06-11
Fix error on proofval continuations constraints

zkronos73:
## Fixes

- Fixed incorrect constraints used to enable continuations for `input_data`, `rom_data`, and `dma`.
- Fixed latent bug in `main` where `stack_enabled=1` was set on witness binaries that lacked the corresponding binary constraint.
- When `stack_enabled` is not configured, the ROM lookup no longer includes the SP-specific columns that are only relevant when the stack is active.
- Fixed several stale/incorrect comments.
- Reorganized `zisk`: moved DMA machines to the end, placing the main machines at the top.


hecmas (approved):
LGTM

PR #1116, opened 2026-06-12
fix(mem-align): constrain step across sub-program rows

hecmas:
## Summary

Fixes a soundness gap in the `MemAlign` air template: the `step` column was never linked across the rows of a sub-program, letting a malicious prover desynchronize the aligned memory accesses from the unaligned operation they implement.

## The bug

`MemAlign` decomposes a non-aligned / sub-word access into one or more aligned 8-byte accesses against the `Mem` SM (the R/W rows) and reconstructs the value returned to Main (the V row). Every other inter-row quantity was tied together — `pc`/`delta_pc` via the ROM, `addr` via `delta_addr`, and `reg` via the `sel * sel_up_to_down` / `sel * sel_down_to_up` transitions — but `step` was a free witness column used only inside the `MEMORY_ID` permutation tuple. It had no transition constraint, no ROM column, and no row-offset reference.

Each aligned access is matched to `Mem` at exactly `(addr, step)`. With the read row's `step` free and decoupled from the value row's `step` (the step Main consumes), a prover could source the aligned word from **any timestamp where `Mem` holds an entry for that address**, not the timestamp of the actual op:

- One-word read at op step `S`, address `X` holding `A` early and `B` after an intervening write: set the read row's step to some `s' < S` where `Mem` legitimately has `(X, s') -> A`. The `Mem` permutation balances, the register
  transitions carry `A`'s bytes into the value row, and the value row proves a stale value to Main at step `S`. The proof verifies.
- Two-word accesses: the two words could be read at unrelated steps; write-back steps were unconstrained relative to the operation.

Neither bus partner can catch this: `Mem` treats each row as an independent access, and Main only sees the value row's step. The linkage has to live in `MemAlign`.

## The fix

Tie `step` across rows so the read/write rows are pinned to the same base step as the value row:

```pil
const expr aligned_wr = wr * (sel_up_to_down + sel_down_to_up);
(1 - reset) * ((step - aligned_wr) - ('step - 'aligned_wr)) === 0;
```

`aligned_wr` is 1 exactly on the two aligned-write rows (a write that is also an assume against `Mem`), which sit one step above the base; it is 0 on read rows and on the value row (which carries `wr=1` in write programs but is a prove row, `sel_up_to_down = sel_down_to_up = 0`). So `base = step - aligned_wr` is invariant within a sub-program, and the constraint enforces `base == 'base` on non-reset rows. `(1 - reset)` starts a fresh base per operation and gates off the cyclic `'step` wrap on row 0.

Completeness is preserved: honest witnesses already follow this layout (read/ value rows at `S`, aligned writes at `S+1`), so no valid trace is rejected — verified row-by-row across all four sub-programs (RV, RWV, RVR, RWVWR); see the truth table added in the comment above the constraint.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1756,6 +1756,9 @@ name = "deranged"
 version = "0.5.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7cd812cc2bc1d69d4764bd80df88b4317eaef9e773c75226407d9bc0876b211c"
+dependencies = [
+ "powerfmt",
+]
 
 [[package]]
 name = "derivative"
@@ -6157,11 +6160,12 @@ dependencies = [
 
 [[package]]
 name = "time"
-version = "0.3.48"
+version = "0.3.47"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fc1aa89044e7786ffb2ec017acb22cb7de5b0be46d0f21aea2b224b8561e5db2"
+checksum = "743bd48c283afc0388f9b8827b976905fb217ad9e647fae3a379a9283c4def2c"
 dependencies = [
  "deranged",
+ "itoa",
  "libc",
  "num-conv",
  "num_threads",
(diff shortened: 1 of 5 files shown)


zkronos73 (dismissed):
LGTM

PR #1124, opened 2026-06-16
fix(zisklib): zisklib bls12 381 msm

han0110:
Updated `msm_complete_safe_bls12_381` and `msm_complete_safe_twist_bls12_381` to do the point membership validation before zero-scalar skip, to be EIP-2537 compliant. Also added the regression test case.

### DIFF
--- a/ziskos/entrypoint/src/zisklib/lib/bls12_381/curve.rs
+++ b/ziskos/entrypoint/src/zisklib/lib/bls12_381/curve.rs
@@ -676,8 +676,8 @@ pub fn msm_complete_safe_bls12_381(
             continue;
         }
 
-        // NOTE: Validation must precede the scalar-zero skip below, since EIP-2537 mandates
-        // rejecting an invalid input point regardless of its scalar.
+        // NOTE: Validation must precede the scalar-zero skip below, since we should reject an
+        // invalid input point regardless of its scalar.
 
         // Verify point coordinates are in the field
         let x: [u64; 6] = point[0..6].try_into().unwrap();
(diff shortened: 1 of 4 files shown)


ziskos/entrypoint/src/zisklib/lib/bls12_381/curve.rs
```diff
@@ -676,15 +676,8 @@ pub fn msm_complete_safe_bls12_381(
...
-        let scalar = reduce_fr_bls12_381(
-            scalar,
-            #[cfg(feature = "hints")]
-            hints,
-        );
-        if is_zero(&scalar) {
-            continue;
-        }
+        // NOTE: Validation must precede the scalar-zero skip below, since EIP-2537 mandates
+        // rejecting an invalid input point regardless of its scalar.
```
hecmas: Could you rewrite the comment to say something more generic as "// NOTE: Validation must precede the scalar-zero skip below, since we should reject an invalid input point regardless of its scalar."? I want the zisklib to not be tailored to any specific application

hecmas (changes requested):

hecmas:
Thanks for this! This is a bug, point validation should occur before skipping.

ziskos/entrypoint/src/zisklib/lib/bls12_381/curve.rs
han0110: Make sense, fixed in https://github.com/0xPolygonHermez/zisk/pull/1124/changes/6a93724c1ae6b054a2c158730cd2d8bcaed3a828

hecmas (approved):
LGTM

PR #1131, opened 2026-06-18
Add BabyJubJub twisted-Edwards point-add precompile

p4u:
## Summary

Adds a circom/circomlib-compatible **BabyJubJub** elliptic-curve point-addition precompile as a dedicated, **isolated AIR** (its own crate, AIR, and lookup table), leaving the shared `arith_eq` state machine untouched so existing ops (secp256k1/bn254/arith256) keep their current proving cost.

- New op-type `BabyJubJub` (opcode `0xEE`) and syscall `SYSCALL_BABYJUBJUB_ADD_ID = 0x81C`
- Host helper over `ark_bn254::Fr` (the BabyJubJub base field is the BN254 scalar field), matching the AIR 1:1
- New `precompiles/babyjubjub` crate: SM, executor, equation generator, planner, instance, bus device, mem-inputs — wired into the executor static SM bundle
- PIL airtemplate + LT table; regenerated `pil/src/pil_helpers/traces.rs`
- Guest example under `examples/babyjubjub` exercising a circomlib known-answer vector

## Design notes for reviewers

- The **complete** twisted-Edwards addition law (`a=168700`, `d=168696`) is decomposed into 7 degree-≤2 polynomial identities over 16×16-bit chunks with carries / 5 scratch columns — the same modelling approach as `arith_eq`'s bn254 curve-add. Because the law is complete, `add(P,P)` doubles correctly with no exceptional-case handling.
- A separate AIR was chosen deliberately: the shared `arith_eq` AIR is hard-capped at 3 equations / 1 scratch column, and widening it would raise proving cost for every existing op. BabyJubJub therefore ships its own crate + AIR + LT table.
- `pil_helpers/traces.rs` is autogenerated from the recompiled PILOUT; AIR_IDs renumber as expected.
- Proving/setup keys (`~/.zisk`) need to be regenerated after this PIL change (standard setup step / CI).

## Test plan

- [x] `cargo fmt --all --check`, `cargo clippy --all-targets` clean
- [x] `cargo check --workspace` builds
- [x] Guest example runs in the emulator and the committed output matches the circomlib reference vector
- [x] `prove --emulator` → `verify` succeeds with the BabyJubJub AIR present in `PROOF INSTANCES SUMMARY` (re-run after setup-key regen)


p4u:
Not yet ready.

PR #1142, opened 2026-06-22
fix(mem-align): soundness in non-aligned op

hecmas:
> Found by @eduadiez.
>
>   The MemAlign component links to the Main SM (the anchor that ties an aligned memory sub-access to a real instruction step) only through the V-row "prove", which is gated by a free witness column sel_prove. No constraint forces sel_prove = 1 on the V row of a program. A malicious prover can therefore emit a structurally-valid program with sel_prove = 0 (an inactive V row), whose R/W rows still consume aligned LOAD/STORE tuples from the memory bus with a free step and a free value. By also adding the matching permutation_proves on the Memory SM side (which the prover controls), the global bus stays balanced — so the only guard that would catch a missing prove never fires. The net effect is a memory write that corresponds to no instruction, at an attacker-chosen address/step/value.

## Summary

The MemAlign state machine reconciles unaligned memory accesses with the aligned-only Memory component. It does this with a single `permutation` against `MEMORY_ID` whose sign selects the direction:

- **assume** rows (sign `-1`): aligned accesses pulled *from* the Memory component.
- **prove** rows (sign `+1`): the reconstructed unaligned value pushed *to* the Main component.

Previously the prove direction was driven by `sel_prove`, a **free witness column** only constrained to be binary and disjoint from the assume selector (`sel_assume = sel_up_to_down + sel_down_to_up`):

```pil
col witness bits(1) sel_prove;
sel_prove * (sel_prove - 1) === 0; // binary
sel_prove * sel_assume   === 0;    // disjoint from assume
...
value[i] === sel_prove * prove_val[i] + sel_assume * assume_val[i];
permutation(MEMORY_ID, [...], sel: sel_prove - sel_assume, sel_is_ternary: 1);
```

Because `sel_prove` was unconstrained on the unaligned "value" rows (where `sel_assume = 0`), a malicious prover could set it to `0`. That row would then be omitted from the permutation (selector `0`) and its `value[i]` would collapse to `0`, so the reconstructed value for an unaligned load/store was never enforced against the Main component. This breaks the consistency guarantee between Main's claimed memory values and the MemAlign computation.

## Fix

Replace the free `sel_prove` witness with `is_non_aligned_op`, a column that is committed but bound by the ROM lookup: it is now part of the flags word checked against MemAlignRom, so its value on every row is forced by the deterministic ROM program for the given pc/line and can no longer be chosen by the prover.
- `mem_align_rom.pil` sets `is_non_aligned_op = 1` exactly on the unaligned value rows (the middle rows of the RWV / RVR / RWVWR access patterns) and packs it into `flags` at bit `CHUNK_NUM + 4`.
- `mem_align.pil` adds the matching `is_non_aligned_op` witness + binary constraint, packs it into flags (matching the ROM), renames `sel_assume` → `is_aligned_op` for clarity, and rewrites the value reconstruction to no longer depend on a free selector:
  ```pil
  value[i] === is_non_aligned_op * non_aligned_val[i] + is_aligned_op * aligned_val[i];
  permutation(MEMORY_ID, [...], sel: is_non_aligned_op - is_aligned_op, sel_is_ternary: 1);
  ```

## Impact

Soundness fix. Closes a gap that allowed a prover to skip the Main-component check for unaligned memory accesses. No change to honest-prover behavior; the PILOUT hash changes, so verifier keys/setup must be regenerated.

### DIFF
--- a/pil/src/pil_helpers/traces.rs
+++ b/pil/src/pil_helpers/traces.rs
@@ -16,7 +16,7 @@ use std::fmt;
 #[allow(dead_code)]
 type FieldExtension<F> = [F; 3];
 
-pub const PILOUT_HASH: &str = "e0eba363372313c7ee099cdbaa73f41d219eac982f53ab31d9edf87b4644a44e";
+pub const PILOUT_HASH: &str = "e2db26c7ce7b878516a8d3a3286915481538d3249a4a172c9c3cc8af4b34a164";
 
 pub const MERKLE_TREE_ARITY: u64 = 4;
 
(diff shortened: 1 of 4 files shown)


zkronos73 (approved):
LGTM

PR #1152, opened 2026-06-26
feat(precompiles): add babyjubjub twisted-edwards point-add precompile

p4u:
Circom BabyJubJub complete point addition over the BN254 scalar field as a dedicated isolated AIR (opcode 0xee, op-type BabyJubJub). Includes the host helper, executor, LT table, PIL equations, guest syscall, executor registration, regenerated PILOUT artifacts, and an end-to-end example validated against the circomlib reference vector.

Aside from the included tests, we have also tested a full flow on our Zisk circuits (https://github.com/vocdoni/davinci-zkvm), which heavily use babyjubjub arithmetic operations. Steps drop from 71,9M to 40,6M (−44%), by using this precompile instead of the zisklib.


hecmas:
@cla-bot check

hecmas (dismissed):
LGTM

hecmas (approved):
LGTM

PR #1166, opened 2026-07-03
Fix rustflags

RogerTaule:
## Problem

When building a ZisK guest, `cargo-zisk` sets `CARGO_ENCODED_RUSTFLAGS` itself to inject `--cfg zisk_guest` and the Zisk linker script. But cargo treats its rustflags sources as **mutually exclusive** — the moment `CARGO_ENCODED_RUSTFLAGS`/`RUSTFLAGS` is set, cargo *ignores* the guest's `.cargo/config.toml` `target.*.rustflags`. So the previous approach silently dropped a guest's own tuning (e.g. the reth guest's `--inline-threshold`) whenever any rustflags env var was present.

Previously the linker script was also injected ad-hoc by callers (see the old `RUSTFLAGS=...-C link-arg=-T...` in `test_diagnostic.sh`), which was fragile and did not preserve config.

## Solution

`ziskbuild` now centralizes guest rustflags resolution in `guest_rustflags()`:

- Resolves the guest's `.cargo/config.toml` rustflags via [`cargo-config2`](https://crates.io/crates/cargo-config2), with the rustflags env vars filtered out so cargo's mutual-exclusion can't suppress config.
- Re-folds the env rustflags back in afterward (in inherit mode) so **both** the program's config tuning and the user's env flags survive, with env keeping cargo's precedence.
- Appends `--cfg zisk_guest` and the linker script (written to a unique temp file, not a predictable path — avoids cross-invocation races and temp-file symlink attacks).

A single `apply_guest_rustflags(&mut Command, …)` helper owns the "set `CARGO_ENCODED_RUSTFLAGS` + scrub the other rustflags env vars" pairing, so the two halves can't drift apart. The CLI `build`/`run` paths and the build-script path all funnel through it.

## Correctness details (matching cargo semantics)

Verified against `cargo-config2`'s resolution source so guest builds match what cargo would do:

- **Mutual exclusion honored.** `env_rustflags()` returns exactly one winning source in cargo's precedence order: `CARGO_ENCODED_RUSTFLAGS` > `RUSTFLAGS` > per-target `CARGO_TARGET_<TRIPLE>_RUSTFLAGS` > `CARGO_BUILD_RUSTFLAGS`. The per-target var is dropped (not combined) when a global source is set — mirroring cargo's `override_target_rustflags`.
- **`CARGO_BUILD_RUSTFLAGS` preserved** as the lowest-precedence env source, so a user's sole `CARGO_BUILD_RUSTFLAGS` still reaches the guest in a direct CLI invocation.
- **Build-script vs CLI split.** Build scripts pass `inherit_env_rustflags=false` (their env vars describe the HOST build, e.g. `-Wl,--export-dynamic`, which would break the guest link); direct CLI invocations pass `true`.
- Uses `cargo_config2::Flags::encode()` for the final value, which validates against `\x1f`-injection instead of silently producing a value cargo would mis-split.

## Testing

- 8 unit tests in `ziskbuild/src/lib.rs` cover: config preservation, empty-flag dropping, host-env isolation in build scripts, inherit-mode env folding, per-target var handling, env-source mutual exclusion, and sole `CARGO_BUILD_RUSTFLAGS`.
- `test_diagnostic.sh` now builds via `cargo-zisk build` instead of open-coding the linker-script `RUSTFLAGS`.
- `cargo clippy` clean on `zisk-build` and `cargo-zisk`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -78,7 +78,7 @@ dependencies = [
  "rand 0.9.4",
  "rapidhash",
  "ruint",
- "rustc-hash 2.1.2",
+ "rustc-hash",
  "secp256k1",
  "serde",
  "sha3",
@@ -540,9 +540,9 @@ checksum = "76a2e8124351fda1ef8aaaa3bbd7ebbcb486bbcd4225aca0aa0d84bb2db8fecb"
 
 [[package]]
 name = "arrayvec"
-version = "0.7.7"
+version = "0.7.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f02882884d3e1bc524fb12c79f107f6ad0e1cfd498c536ffb494301740995dfe"
+checksum = "d3fb67a6e08acf24fdeccbac2cb6ac4305825bd1f117462e0e6f2f193345ad56"
 
 [[package]]
 name = "asm-runner"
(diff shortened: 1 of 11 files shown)


zkronos73 (approved):
LGTM

PR #1178, opened 2026-07-08
add sections .sbss to zisk linker script

zkronos73:
Add C++ used section .sbss.* inside bss sections of zisk linker script


RogerTaule (approved):
LGTM

PR #1196, opened 2026-07-16
feat(blake2): Soudness bug and 3x denser arithmetization for Blake2

hecmas:
## Summary

Re-arithmetizes the `Blake2br` precompile from 24 rows per round call down to **8 rows (one full G function per row)**, tripling the number of BLAKE2b rounds provable per AIR instance: with `N = 2^18`, capacity goes from 10,922 to **32,768 ops** (8 | 2^18, so the trace also no longer loses a reserved tail cycle). The syscall ABI, bus payloads, and memory access map are unchanged — callers and the mem planner are untouched.

## Arithmetization changes (`precompiles/blake2/pil/blake2br.pil`)

- **One G per row.** Each row takes `(va, vb, vc, vd)` plus both message words `x`, `y` and computes the full G function through in-row intermediate columns (`va'`, `vd'`, `vc'`, `vb'`, `va''`, `vd''`, `vc''`). Rows 0–3 are the column mixing, rows 4–7 the diagonal mixing; transition constraints route column outputs into diagonal inputs with per-variable rotation offsets.
- **XOR + rotations via a byte table** instead of per-bit boolean decomposition (the old row carried 128 one-bit columns for `vb`/`vd`). A new virtual table `Blake2brTable` (2^16 rows, `A ^ B = C`) proves all XORs byte-by-byte; the `>>> 32/24/16` rotations are free byte re-indexings of the lookup outputs. `>>> 63` is handled as `<<< 1` with two boolean carry bits (sound by the transition equality binding the result into a `< 2^32` limb).
- **Additions mod 2^64** keep the 32-bit-limb CRT check (Plonky3-style), with the helpers reworked to take the result last (`d == a + b + c`).
- **Message permutation, two words per row.** `x`/`y` hold the memory-ordered words (`m[2k]`, `m[2k+1]` at clock `k`), `xs`/`ys` the SIGMA-permuted words consumed by G. The permutation argument now carries two tuples per side per row (`MSG_IDX`/`MSG_IDX+1` vs `sigma_idx[0..2]`), tagged by `op_step` and gated by `in_use` so unused cycles contribute nothing.

## Memory interface

Same 51 memory ops per call, rescheduled onto 7 ports so every port's value is the current row's columns (no shifted value references):

|                | clk 0 | clk 1  | clk 2  | clk 3 | clk 4  | clk 5  | clk 6  | clk 7  |
|----------------|-------|--------|--------|-------|--------|--------|--------|--------|
| port 0 (`va`)  | R v0  | R v1   | R v2   | R v3  | W v'0  | W v'1  | W v'2  | W v'3  |
| port 1 (`vb`)  | R v4  | R v5   | R v6   | R v7  | W v'5  | W v'6  | W v'7  | W v'4  |
| port 2 (`vc`)  | R v8  | R v9   | R v10  | R v11 | W v'10 | W v'11 | W v'8  | W v'9  |
| port 3 (`vd`)  | R v12 | R v13  | R v14  | R v15 | W v'15 | W v'12 | W v'13 | W v'14 |
| port 4 (`x`)   | R m0  | R m2   | R m4   | R m6  | R m8   | R m10  | R m12  | R m14  |
| port 5 (`y`)   | R m1  | R m3   | R m5   | R m7  | R m9   | R m11  | R m13  | R m15  |
| port 6 (params)| R idx | R ptr0 | R ptr1 |       |        |        |        |        |

The diagonal write rotation (`offset_shift: port`) matches the diagonal G output routing; the committed `op_step` broadcast doubles as the memory `main_step`.

## Witness (`precompiles/blake2/src/`)

- `process_input` rewritten for the 8-row layout: fills the per-row G intermediates, both message-word views, and collects multiplicities for the 16-bit range checks (16 limbs/row) and the new XOR table (32 lookups/row) in one pass.
- Multiplicities are aggregated with rayon `fold`/`reduce` instead of collecting one 64K array per op.
- New `Blake2brTableSM` registering the XOR table rows via the std virtual-table API; padding rows account for their zero-valued range checks and `(0,0,0)` lookups.

## Misc

- `zisklib::blake2b_compress`: branchless working-vector initialization.
- `BLAKE2BR_TABLE_ID = 128` in `opids.pil`; `virtual Blake2brTable` registered in `zisk.pil`; regenerated `pil_helpers` trace bindings.
- Requires a setup regeneration (new columns, constraints, and pilout hash).


zkronos73 (approved):
LGTM

PR #1213, opened 2026-07-23
fix(aritheq384): Soudness in ArithEq384

hecmas:
> Found by [zkao](https://zkao.io/).

### Problem

`ArithEq384` uses cycles of `CLOCKS = 24` rows, which does not divide `N = 2^20`. The trace therefore has a non-usable tail, and the original layout had `CLK_0 = 0` on all of its rows.

When an instance was **fully occupied**, the padding rows started right after the last real cycle with no `CLK_0 = 1` row in between, so the latch-style constraints never released:
- `(1 - CLK_0) * (sel_op[i] - 'sel_op[i]) === 0` forced the last operation's selectors to propagate through every padding row.
- `x_are_different === 'x_are_different * (1 - CLK_0) + x_chunk_different` forced `x_are_different` to propagate as well.

All-zero padding rows therefore violated these constraints. Copying the previous `sel_op` / `x_are_different` values into the padding is not a valid fix either: non-zero selectors keep `sel_check_x_lt_prime` / `sel_check_y_lt_prime` active in the padding, firing `lookup_assumes` against `ARITH_EQ_LT_TABLE_ID` with keys that are not matched by the table (e.g. the first padding row produces the transition `'x3_lt = 1 → x3_lt = 0` with `delta_x3 = 0`, which requires `delta > 0` in the table).

### Fix

Instead of releasing the latches implicitly (an earlier version of this PR inserted a ghost `CLK_0 = 1` at the start of the tail, sacrificing one cycle), the padding behavior is now explicit:
- The base clock — renamed `CLK_0` → `FIRST_CLK` — is zero on the whole tail, and `NUM_OPS` no longer reserves a cycle, so the tail shrinks to `N % CLOCKS = 16` rows and **each instance fits one more operation**:
  ```
  col fixed air.FIRST_CLK = [[1, 0:(CLOCKS-1)]:NUM_OPS, 0...];
  ```
- A new fixed expression `SECOND_CLK_TO_LAST = CLK[1] + ... + CLK[CLOCKS-1]` covers clocks 1 to 23 of real cycles and is zero on the tail. It gates both latches, so they release at every cycle start *and* at the padding boundary:
  ```
  SECOND_CLK_TO_LAST * (sel_op[i] - 'sel_op[i]) === 0;
  x_are_different === 'x_are_different * SECOND_CLK_TO_LAST + x_chunk_different;
  ```
- Padding rows are now explicitly constrained rather than merely unconstrained-but-zeroable:
  ```
  (1 - FIRST_CLK - SECOND_CLK_TO_LAST) * in_use === 0;
  ```
  The gate is exactly 1 on tail rows, and since `in_use` is a sum of six boolean-constrained selectors (no field wrap possible), this pins every `sel_op[i]` to 0 there. Nothing can fire from the padding: operation bus (`in_use_clk0 = 0`), memory bus (`mem_sel = 0`), and lt-table lookups (`sel_check_*_lt_prime = 0`) are all provably dead on the tail.
- The `sel_op_clk0` projection collapses to a single definition:
  ```
  sel_op_clk0[i] <== sel_op[i] * FIRST_CLK;
  ```
  which subsumes three constraints per selector (binarity, zero off the first clock, equality on it); `in_use_clk0`'s binarity is implied as well. Net: 12 fewer constraints.

All-zero padding rows satisfy every constraint, in both partially- and fully-occupied instances, and — unlike the ghost-cycle approach — no `CLK[i] = (i)'FIRST_CLK` shift ever lands on an active row across the trace boundary, so there is no wrap-around case to reason about.


zkronos73 (approved):
LGTM

PR #1217, opened 2026-07-24
fix(mem-align): Soundness bug in Memory Align SM

### LINKED ISSUE #1188: Security issue in ZisK's memory-alignment state machine (`MemAlign`)
I have identified a soundness issue in ZisK's memory-alignment state machine (`MemAlign`) affecting zero-extending sub-word loads (`lhu`, `lwu`), and I have a faithful, executed proof of concept at the constraint level. Before contacting you I re-verified the finding against the **current `main` branch** (`8447ee07`, 2026-07-03), because I noticed two adjacent MemAlign gaps were fixed on `main` in June — those fixes do **not** cover this one, and it is still present.

The ZisK source is public, so I'm including the full technical detail below. I can also share the runnable PoC on request.

### Summary

* **Target:** ZisK (`0xPolygonHermez/zisk`).
* **Affected component:** `state-machines/mem/pil/mem_align.pil` (the general `MemAlign` SM) in combination with the `copy_b` operation in `state-machines/main/pil/main.pil`.
* **Affected versions:** current `main` (`8447ee07`, 2026-07-03) and, as far as I can tell, **every published release** (through `v1.0.0-alpha`). The vulnerable value-reconstruction logic has been present since the memory SM was introduced. I verified `main` by direct source inspection of the upstream files; the executed PoC was run against the `MemAlign` constraint set from the pinned tree `6182c8be9` (2026-06-04), whose relevant lines are byte-identical to `main`.
* **Vulnerability class:** Soundness — memory/load consistency. A malicious prover can make any `lhu` (upper 48 bits) or `lwu` (upper 32 bits) load return an **arbitrary chosen value in the high bits** — a value that was never in memory — and the verifier accepts.

### Mechanism (`main`, `8447ee07`)

For any sub-word access (`width < 8`), ZisK proves the load through the general `MemAlign` SM. The prove/`[V]` row reconstructs the delivered `value[]` from **all eight** register byte-lanes (`mem_align.pil`, current line numbers):

```
// mem_align.pil:198-208 — value is reconstructed from ALL lanes, rotated by the selector
for (int _offset = 0; _offset < CHUNK_NUM; _offset++) {
    for (int rc_index = 0; rc_index < RC; rc_index++) {
        expr _tmp = 0; int base = 1;
        for (int ichunk = 0; ichunk < CHUNKS_BY_RC; ichunk++) {
            _tmp += reg[(_offset + rc_index * CHUNKS_BY_RC + ichunk) % CHUNK_NUM] * base;
            base *= 256;
        }
        prove_val[rc_index] += sel[_offset] * _tmp;
    }
}
// mem_align.pil:211-213 — no width mask anywhere
col witness bits(32) value[RC];
for (int i = 0; i < RC; i++) {
    value[i] === sel_prove * prove_val[i] + sel_assume * assume_val[i];
}
```

The only thing that ties the `[V]`-row register lanes back to the value actually read from memory is the R→V transition constraint, and it is **gated by the R-row selector `sel[i]`**, which is `1` only for the in-window lanes `[offset, offset+width)`:

```
// mem_align.pil:113-116
range_check(reg[i], 0, 2**CHUNK_BITS-1);          // out-of-window lanes: byte-range-checked ONLY
(reg[i]' - reg[i]) * sel[i] * sel_up_to_down === 0; // pins in-window lanes; out-of-window lanes are FREE
```

So the byte-lanes **outside** the loaded width are unconstrained except for a 0–255 range check. `width` does appear on the memory permutation, but only as a *matched* bus field — it never masks `value`:

```
// mem_align.pil:215
permutation(MEMORY_ID, expressions: [ ..., width, ...value], sel: sel_prove - sel_assume, sel_is_ternary: 1);
```

Downstream, `lhu`/`lwu` are translated to the internal `copy_b` operation (`core/src/riscv2zisk_context.rs:227` and `:229`), and `copy_b` writes the memory value straight to the destination register with **no width mask**:

```
// main.pil:137  — [is_external_op == 0 && op == 1] ==> c = b   (copy_b)
// main.pil:396
(1 - is_external_op) * op * (b[index] - c[index]) === 0;   // c === b for BOTH 64-bit limbs, index = 0,1
// main.pil:311-312 — c is then written out via store_value → destination register
```

**Consequence.** For an `lwu`/`lhu`, the honest prover sets the out-of-window high lanes to `0` (zero-extension) — but nothing *constrains* them to `0`. A malicious prover sets those lanes to arbitrary bytes; `value` carries them; `copy_b` copies them into `rd`. The register receives a value the honest program can never produce, and every downstream computation and public output built on it is forged while the verifier accepts.

### Why `lhu`/`lwu` specifically (scope)

* **`lb`/`lh`/`lw`** are translated to `signextend_b/h/w` (`riscv2zisk_context.rs:224/226/228`) — a separate operation that re-derives the high bytes from the sign bit, so the free high lanes are overwritten → **not exploitable**.
* **`lbu`** (width 1) is routed to the dedicated `MemAlignByte` SM (`mem_align_byte.pil`), which composes the result byte-by-byte rather than from free lanes → **not exploitable**.
* **`lhu` (width 2)** and **`lwu` (width 4)** are translated to `copy_b` (`riscv2zisk_context.rs:227/229`) and go through the **general `MemAlign`** path above → **exploitable**. `lhu` → upper 48 bits forgeable; `lwu` → upper 32 bits forgeable.

### Fix

In `MemAlign`, for zero-extending loads force the lanes outside `[offset, offset+width)` to `0` before they enter the value reconstruction (equivalently, mask `value` by `width` prior to the `MEMORY_ID` permutation), so the committed loaded value cannot carry unread high bytes. Alternatively, have the `copy_b` load path mask its result by the load width in `main.pil`. Either change closes the gap; the `MemAlign`-side mask matches how `MemAlignByte` already constrains its high bytes.

---

### Secondary, liveness

1. **`divw` with `INT_MIN / -1` — executor conformance + liveness.** `op_div_w` (`core/src/zisk_ops.rs`, opcode `0xbe`) computes `(((a as i32) as i64) / ((b as i32) as i64)) as u64`, which for `-2^31 / -1` yields `0x0000_0000_8000_0000` (zero-extended) instead of the RISC-V-mandated `0xFFFF_FFFF_8000_0000` (sign-extended). This is **not** exploitable for forgery: the committed arith PIL binds the correct sign-extended result (`arith.pil:268`, `bus_res1 <== sext * 0xFFFF_FFFF + (1 - m32) * bus_res1_64`, with `sext` forced to 1 for this row by the arith table), so an honest prover following the buggy executor produces a Main↔Arith bus that is **unsatisfiable → the proof cannot be generated** (a liveness/DoS issue on any program that hits this operand pair). Reachability is low because safe Rust guards division overflow and never emits `divw INT_MIN/-1`; it only bites hand-written or adversarial assembly. Cross-checked as ZisK-specific (Jolt, lambdaworks-vm, and QEMU all sign-extend correctly). Suggested fix: `((a as i32).wrapping_div(b as i32) as i64) as u64`.

hecmas:
> Originally reported by @liyue-cs in [issue #1188](https://github.com/0xPolygonHermez/zisk/issues/1188).
> Independently reproduced by @codygunton in [PR](https://github.com/codygunton/zisk/pull/10).

### Summary

Closes a soundness gap in the general `MemAlign` machine that let a prover forge the out-of-window bytes of a narrow unsigned load (`lwu`, `lhu`, `lbu` routed through the general machine), so the returned `rd` could contain bytes that were never in memory. Adds three width-category selectors that force those bytes to zero on load prove rows.

### The vulnerability

On a `MemAlign` prove row the returned value is reconstructed from the 8 registers via the one-hot `sel` rotation. Only the in-window lanes `[offset, offset+width)` are pinned — the read→value continuity gate (`(reg[i]' - reg[i]) · sel[i] · sel_up_to_down`) ties only the lanes the adjacent aligned read marks, and the `MemAlignRom` lookup pins row structure but no data lane. Every byte position `≥ width` therefore comes from a register that is constrained *only* to be a byte.

For a narrow load the high bytes must be zero (zero-extension), but nothing enforced it. Since Main consumes the loaded value without masking (`lwu → copyb → c = b → rd`), a forged high byte flows straight into `rd`. Concretely, a width-4 offset-0 `lwu` could return `reg[4..7]` as the high limb — e.g. `rd = 0x0000_0001_0000_0000` — and both `verify-constraints` and a full proof accepted it.

The gap spans every **load** prove row:
- **RV** (single-word), widths 1/2/4
- **RVR** (two-word), the narrow cases `(offset 5, w4)`, `(offset 6, w4)`, `(offset 7, w2)`, `(offset 7, w4)`

For width 2 the low limb is affected too (bits 16–31 of `value[0]` come from out-of-window lanes), and for width 1 through the general machine all but the loaded byte are forgeable — so the fix pins the low-limb upper bytes as well, not just the high limb.

### Why loads only

Store prove rows (`RWV`, `RWVWR`, `wr = 1`) are **not** affected: their value is an input pinned by Main's `c` (register bus), and only the in-window low bytes reach memory via the write mask. So the fix is gated to `wr = 0`; applying it to stores would reject honest narrow stores whose source register holds nonzero high bits.

### The fix

Three width-category selectors, emitted by the ROM (fixed columns, packed into the existing `FLAGS`, pinned through the `MEMORY_ALIGN_ROM_ID` lookup) and set to `1` only on a load prove row of the given width:

- `sel_w_lt8` — `is_non_aligned_op ∧ ¬wr ∧ width < 8`
- `sel_w_lt4` — `… ∧ width < 4`
- `sel_w_lt2` — `… ∧ width < 2`

New constraints (all degree ≤ 3, matching the AIR's ceiling), where `vbyte[k] = Σ_offset sel[offset]·reg[(offset+k) % CHUNK_NUM]` is value byte `k`:

```
sel_w_lt8 * value[RC-1] === 0;   // width < 8  -> high limb (bytes 4..7) zero
sel_w_lt4 * vbyte[3]    === 0;   // width < 4  -> byte 3 zero
sel_w_lt4 * vbyte[2]    === 0;   // width < 4  -> byte 2 zero
sel_w_lt2 * vbyte[1]    === 0;   // width == 1 -> byte 1 zero
```

The `non_aligned_val` reconstruction and the `MEMORY_ID` permutation are left untouched, so nothing on the bus-facing path changes. The honest witness already fills out-of-window lanes from a zero-extended value, so honest RV/RVR traces satisfy the new constraints unchanged; only forged rows are rejected.

### Secondary change: dual-byte register range checks

The 8 per-register `range_check(reg[i], 0, 255)` calls are replaced by 4 `range_dual_byte(reg[i], reg[i+1])` lookups against the existing `DualByte` table, halving the range-check bus interactions per row. Same `[0,255]` bound per register (the two `DualByte` lanes are independent). Assumes `CHUNK_BITS == 8`, which is the machine's configuration.


zkronos73 (approved):
LGTM

PR #1219, opened 2026-07-27
Fix/bug 1.1.0 alpha

RogerTaule:


xavi-pinsach (approved):
The Rust part looks good to me. Pending to check the C part.

fractasy (approved):
Agreed with the assembly change.

PR #1220, opened 2026-07-27
Fix (coordinator): validate worker task responses against job phase and assignment (audits issue #36)

xavi-pinsach:
Hardens the coordinator's handling of worker task responses. Reported as 0xPolygonHermez/audits#36, though the fixes stand on their own as protocol robustness: responses were routed on payload variant alone, so an out-of-order or buggy worker could drive a job through a phase it was not in — including completing it.

### Changes

**`817bc30c0` — record the aggregation task in-flight before dispatching**

`handle_proofs_completion` released the job lock, sent the task to the recurser, then recorded it in `agg_task_inflight`. The recurser can reply inside that window. `dispatch_next_agg_task` already recorded before sending, so the two dispatch sites disagreed; this aligns them and clears the slot if the send fails.

**`bd4fd78eb` — bind responses to the job's current phase**

Adds `validate_payload_phase`, an exhaustive payload-variant → allowed-phase table applied before dispatch. Two asymmetries in the table are load-bearing:

- `Proofs` is accepted in `Recurse` as well as `Prove`, because `resolve_recurser_assignment` flips the job to `Recurse` as soon as the *first* worker finishes phase 2 — the rest legitimately report in afterwards. Requiring `Prove` alone would break every multi-worker job.
- Execution-only jobs run in `Running(Contributions)`; `JobPhase::Execution` is only ever a `phase_timings` key, never a job state. The two phase-1 variants are therefore told apart by `execution_only`, not by phase.

Also has `validate_and_update_heartbeat` return the job it already looked up, so the gate reuses that handle instead of repeating the lookup and runs unconditionally rather than behind an `Option` that could skip it.

**`7bb6b3f75` — accept final proofs only from the designated recurser**

`handle_recurser_completion` checked nothing beyond job assignment. It now requires the sender to be `job.agg_worker_id`, an aggregation task to be in flight, and — for a non-empty proof — that task to be the one carrying `all_done`. The first two run before the empty-proof branch, since that branch mutates the aggregation queue.

This also removes a panic: `job.agg_worker_id.as_ref().unwrap()` aborted the connection task whenever a `FinalProof` arrived before a recurser was assigned (`agg_worker_id` is only set during phase 2), reachable by any worker sending out of order.

**`afe503ace` — require agreement on execution-only public outputs**

Execution-only jobs return public outputs with no proof involved. Phase 1 already cross-checks `publics` across workers; this path did not, and served `results.values().next()` — arbitrary `HashMap` order — so one disagreeing worker could decide the result. Now unanimity is required and the outputs are taken from that same check. Skipped in simulation mode, where `job.workers` lists one physical worker N times.

### DIFF
--- a/distributed/crates/coordinator/src/coordinator.rs
+++ b/distributed/crates/coordinator/src/coordinator.rs
@@ -2232,6 +2232,94 @@ mod tests {
         assert!(job.proof.is_none(), "no proof may be stored from a rejected response");
     }
 
+    /// Only the designated recurser may report aggregation results. Any other
+    /// assigned worker submitting a `FinalProof` in `Recurse` is rejected.
+    #[tokio::test]
+    async fn test_final_proof_rejected_from_non_recurser() {
+        let (coordinator, workers, job_id) =
+            setup_coordinator_with_job(2, JobPhase::Recurse, |_| {}).await;
+        let (w0_id, w1_id) = (workers[0].0.clone(), workers[1].0.clone());
+
+        // w0 is the recurser with a final task outstanding; w1 tries to answer.
+        {
+            let entry = coordinator.jobs.read().await.get(&job_id).cloned().unwrap();
+            let mut job = entry.write().await;
+            job.agg_worker_id = Some(w0_id.clone());
+            job.agg_task_inflight = Some(zisk_cluster_common::PendingAggTask {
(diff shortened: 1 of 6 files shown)


RogerTaule (approved):
LGTM

PR #1228, opened 2026-07-29
fix soundness bug with arith state-machine on signed division

zkronos73:
> Fixes the signed DIV quotient-sign forgery reported by @codygunton in [codygunton/zisk#12](https://github.com/codygunton/zisk/pull/12).
> Also closes the sibling signed-REM remainder-sign bug independently reported by @pjordanandrsn [GHSA-qjgg-fcj3-p9x3](https://github.com/0xPolygonHermez/zisk/security/advisories/GHSA-qjgg-fcj3-p9x3).

## Summary

The motivation is that bug fix; the rest of the change is work in the same area that it made
convenient.

Arith encodes signs as single bits: `na`, `nb`, `np`, `nr` are 1 when the value is negative and 0
otherwise. The catch is that "otherwise" means **positive or zero** — zero is the one value that
belongs to both halves, and a sign bit cannot tell the two apart. The sign rules of a division
(`sign(quotient) = sign(dividend) XOR sign(divisor)`, `sign(remainder) = sign(dividend)`) hold for
non-zero values and have an exception exactly at zero, so with only the sign bits the constraints
written as a table had to **leave the exceptional combinations open** — and once open they were also
accepted with non-zero values, which is the bug.

`result_is_zero` and `remainder_is_zero` name that exception. The general rule is now pinned in the
truth tables, the exception is admitted only when the flag is set, and the flag in turn forces the
value to really be zero. Nothing is left open.

The two extra columns are also why the `multiplicity` selector could be dropped in exchange for an
air value, and the same pass reduced the range-id layout and tightened the carry range check.

## Two new flags for the zero edge cases

`div_overflow_mul_rz` was overloaded: it meant `div_overflow` on divisions and "the product is zero"
on multiplications, and there was no way at all to say "the remainder is zero". It is now three
independent flags: `div_overflow`, `result_is_zero` and `remainder_is_zero`.

That is what closes the bug reported in the PR linked above. The sign relation of a division is
`sign(quotient) = sign(dividend) XOR sign(divisor)`, **except when the quotient is zero**, and
likewise `sign(remainder) = sign(dividend)` except when the remainder is zero. Without a flag to mark
those exceptions the table had to admit the exceptional sign combinations unconditionally — and
nothing forced the value to actually be zero. Measured on the old table: **22 of its 80 rows were
reachable only when the quotient or the remainder is zero, with no flag to pin it.** For instance the
`div` row with `na=0 nb=1 np=0 nr=0` (non-negative quotient, negative divisor, non-negative dividend)
is consistent only with a zero quotient, yet a non-zero quotient was accepted there.

Now the truth tables pin it directly:

```
neg_quotient  = (neg_dividend XOR neg_divisor) AND NOT quotient_is_zero
neg_remainder =  neg_dividend                  AND NOT remainder_is_zero
```

and `result_is_zero` / `remainder_is_zero` force the corresponding value to zero in the AIR.

## A more intuitive table

The per-case validity filters were replaced, where possible, by these two truth tables written out
row by row, which is far easier to read against the RISC-V semantics than a list of `continue`s. The
filters that became redundant are commented out with the reason, and the ones that remain each reject
something real. Two aliases (`neg_dividend`, `neg_divisor`, `neg_quotient`, `neg_remainder`) make the
division rows say what they mean: for a division, `np` is the sign of `c`, the **dividend**, not of the
quotient — the single most confusing thing in the old text.

The table goes from 80 to 124 rows. All of the growth is the two new flags refining existing states;
every one of the 124 rows is reachable by a real operation, which is now machine-checked.

## Carry range

Tightened from ±983 040 to **±262 142**, the attainable bound. Column `k` of `eq[]` holds up to `n_k`
products of two 16-bit chunks with `n = [1,2,3,4,3,2,1]`, so `|carry[k]| <= n_k * 2^16` and the binding
index is `carry[3]`. `cargo run --release --bin arith_carry_range` recomputes it exactly by vertex
enumeration, which is valid because each `eq[k]` is multilinear in the chunk variables.

## Range-id layout: 43 slots to 20

A range id carries the range type of four chunks at once, read through four offsets into one shared
slot sequence. The offsets were `0 / 26 / 17 / 9`, so the four views barely overlapped and the
sequence needed 43 slots. With **consecutive offsets `0 / 1 / 2 / 3`** they overlap maximally and the
same `4 x 17 = 68` requirements fit in **20 slots**. `ArithRangeTable` chunk rows go from 2 228 224 to
1 048 576.

This is the minimum, not just an improvement: each of the four views needs 17 distinct slots of which
exactly 11 must be FULL, 3 POS and 3 NEG, so the sequence needs at least 11 FULL slots; and four
distinct offsets spanning `s` force `17 + s >= 20` slots. The cost in 2^15 units is
`#slots + #FULL >= 31`; 31 is unreachable by exhaustive search and any span above 3 already needs 32.
The argument is in the `COL COMPRESSION` note of `pil/arith_range_table.pil`.

Note that the all-FULL range id moves from 0 to 3, so `ARITH_RANGE_16_BITS` moves with it.

## Selector column removed, replaced by an air value

The two new flags add two columns, so to keep the width the `multiplicity` selector of
`proves_operation` was dropped. Every row now proves an operation unconditionally, and the padding
rows are cancelled off the bus with an air value:

```pil
airval padding_size;
assumes_padding_operation(op: OP_MULU, a:[0,0], b:[0,0], c:[0,0], flag:0, padding_size:);
```

Padding rows are therefore a real, trivial operation: **`mulu(0, 0) = (0, 0)`**. They are built from an
`ArithOperation` rather than hand-written, so the trace columns and the table multiplicities come from
the same source. Same pattern as `BinaryAdd` and `BinaryExtension`.

## Diagnostic coverage

`test-artifacts/programs/diagnostic/src/zisk_arith.rs` is new: it runs all 13 RISC-V M-extension
instructions that land on `Arith` over all ordered pairs of 14 representative values
(0, 1, 2, `MAX/2 ± {0,1,2}`, `MAX - {0,1,2}` and their two's complement negations — the set is closed
under negation), checking each against the `ops_core` reference. 2548 checks; for the `_W`
instructions it verifies the sign extension of the full register as well as the low half. It reaches
122 of the 124 table rows.

## Reviewing the table change

`state-machines/arith/docs/arith_table.txt` is the table decoded to a canonical form: flags named,
range ids expanded into the range **type** of each of the eight constrained chunks. Comparing it with
the base is what shows the semantic delta, since neither flag bit positions nor range id numbers are
comparable across this change:

```
git show <base>:state-machines/arith/src/arith_table_data.rs > /tmp/old.rs
cargo run --release --bin arith_table_decode_gen -- /tmp/old.rs --legacy --coarse > /tmp/old.txt
cargo run --release --bin arith_table_decode_gen -- --stdout --coarse            > /tmp/new.txt
diff /tmp/old.txt /tmp/new.txt        # empty
```

`--coarse` keeps the part both encodings can express. **The diff is empty**: the same 74 states, and
the same range type on all eight chunks for each of them. No state added, none removed, no range
constraint changed — the layout reduction is representation only. Without `--coarse` the diff shows
the refinement: 46 states unchanged, 28 split `1 -> 2` or `1 -> 4` by the new flags, 0 shrank.

Attached, so the diff can be read without running anything:

| file | contents |
| --- | --- |
| [arith_table_before.txt](https://github.com/user-attachments/files/30525625/arith_table_before.txt)| the 80 old rows, decoded |
| [arith_table_after.txt](https://github.com/user-attachments/files/30525624/arith_table_after.txt) | the 124 new rows, decoded |
| [arith_table_before_coarse.txt](https://github.com/user-attachments/files/30525623/arith_table_before_coarse.txt) | the 74 states, old |
| [arith_table_after_coarse.txt](https://github.com/user-attachments/files/30525622/arith_table_after_coarse.txt)| the 74 states, new — identical to the previous file |

## Tests

`cargo test -p sm-arith` covers, beyond the existing suite:

* `table_coverage_tests` — every table row is produced by some real operation and every operation has
  a row (145 656 operations, all 124 rows).
* `range_layout_tests` — `RANGE_PATTERN`, `RANGES`, `OFFSETS`, `CARRY_BASE` and `get_range_name` agree,
  and the 20-slot sequence really realises the 17 required tuples.
* `padding_row_tests` — the padding operation has a table row, every range id it uses admits zero, and
  nothing beyond the explicitly written columns needs setting.
* `committed_decoding_is_up_to_date` — `docs/arith_table.txt` matches the current table.


hecmas (approved):
LGTM

PR #1236, opened 2026-08-03
Remove redundant copies of job input on the coordinator and worker

RogerTaule:
The inline-input path copied each job's input far more than it needed to. For a 32.5 MB block across 8 workers that was ~520 MB of memcpy per job, much of it serial in front of the Phase 1 fan-out.

Coordinator:
- Drop the hex round-trip. InputsModeDto::InputsData and HintsModeDto::HintsData now hold raw bytes rather than a hex String, removing an encode on submit and a decode on dispatch (~137 ms on a 32.5 MB input) and halving what a job retains in memory.
- Generate Bytes for the payload fields that fan out to every worker, so per-worker clones are refcount bumps and prost decodes a slice of the wire buffer. The wire format is unchanged.
- Clone inside each dispatch future rather than at the iterator level: buffer_unordered drains the whole iterator before polling any of it, so the previous shape ran every copy before the first message was queued.
- Convert relay chunks once outside the per-worker loop.
- Take the input by value in domain_input_to_dto / domain_hints_to_dto.

Worker:
- ZiskStdin held the buffer twice: `data`, plus a Cursor over a clone of it, with every write appended to both. It now keeps one buffer and a read position. read_raw commits that position only on success, so a short read no longer leaves the frame reader desynced.
- Add ZiskStdin::with_data so read-only callers stop cloning, and use it where the ASM input is written into shmem.
- Skip the per-job MPI serialization when n_processes == 1. MpiCtx::broadcast already no-ops at one rank, so only the borsh pass over the payload was being paid. Adds n_processes() alongside world_rank() on the Prover trait.

Borsh impls for InputSourceDto and HintsSourceDto are hand-written, since Bytes has none; the encoding matches what the derive produced, so the MPI format is unchanged. Tests cover that layout and the ZiskStdin framing.


rickb80 (approved):
LGTM

PR #1248, opened 2026-08-10
Overlap Main witness computation with ASM minimal-trace execution

xavi-pinsach:
**Main AIR instances** are now planned, assigned and marked witness-ready **during** the ASM minimal-trace run instead of after it.
As soon as the emulator has streamed enough chunks to fill one Main segment, that instance is handed to proofman — so its witness computation, and the GPU streaming-slot commit that follows, overlap the remaining emulation and the gpu-mops window instead of queuing behind `await_mem_plans`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -628,6 +628,7 @@ dependencies = [
  "mem-common",
  "mem-planner-cpp",
  "named-sem",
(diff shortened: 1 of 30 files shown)


rickb80 (approved):
LGTM

PR #1249, opened 2026-08-10
Miscelaneous

RogerTaule:
- Coordinator setup-failure handling, dev-CLI reporting, keccak inlining coordinator: a worker whose setup failed was flipped to Ready anyway (`ack.success` gated only the logging), so the dispatcher handed tasks to workers with no artifacts for the program. Park them at Idle instead — out of dispatch, still eligible for a setup retry, and not counted in the whole-fleet capacity floor, so jobs run at reduced width instead of wedging. Same fix in the recurser ack path, plus a regression test.
- dev prove: print timings and the summary even with no final proof, so `--no-aggregation` runs still report a total.
- dev stats: add `--gpu` / `--cpu-mops` so the timings match what the proving pipeline does, and derive `air_name` from the generated AIR_NAMES table instead of a drifted hand-written match.
- keccak_f utils: inline the state conversion helpers, downgrade the bit-position bounds check to a debug assert.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -631,7 +631,7 @@ dependencies = [
  "proofman-starks-lib-c",
  "proofman-util",
  "rayon",
- "thiserror 2.0.19",
+ "thiserror 2.0.20",
  "tracing",
  "zisk-common",
  "zisk-core",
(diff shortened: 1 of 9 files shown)


xavi-pinsach (approved):
LGTM

PR #1251, opened 2026-08-11
Fix bugs in the DMA planner

zkronos73:
Three independent bugs in the counters → plan path of the DMA precompile, all found while
reviewing the instance-selection strategy. Two of them can invalidate a run; the third is a
memory/robustness problem. Plus the first unit tests for the crate.

No behavioural change is intended beyond the fixes: the strategy still routes the same
operations to the same airs, and the PIL is untouched.

## 1. Planner panics when the full air's rows are an exact multiple of the capacity

`calculate_dma_strategy` computes how many rows are still free in the tail of the last `Dma`
(resp. `DmaPrePost`) instance, so it can fold the `memcpy`/`inputcpy` remainders in there and
save whole instances:

```rust
let remain_dma = rows_full % rows_x_full_instance;
let available_on_dma = if rows_full == 0 { 0 } else { rows_x_full_instance - remain_dma };
```

When `rows_full` (= `memset` + `memcmp` rows) is a non-zero exact multiple of the capacity,
`remain_dma == 0` and this reports `rows_x_full_instance` free rows — a whole free instance —
when the last instance is in fact completely full.

The planner then decrements `info.memcpy` and routes the remainder into the full air, which now
receives more rows than the instances it asked for can hold, and
`DmaInstancesBuilder::open_new_instance` aborts:

```
[dma_full] Too many instances 1 max: 1, cannot create more
```

Deterministic crash at planning time, so no invalid proof — but it depends only on a total row
count landing exactly on a multiple of 2^21, i.e. it is input-dependent and would show up as a
sporadic failure on an otherwise valid program. Affects both `Dma` and `DmaPrePost`.

Fixed by taking the remainder modulo the capacity, which yields 0 both for an exact multiple and
for `rows_full == 0`, subsuming the previous special case:

```rust
let available_on_dma = (rows_x_full_instance - remain_dma) % rows_x_full_instance;
```

## 2. `skip` charged once per instance instead of once per batch

In `DmaInstancesBuilder::add_op_rows`, `skip` is the number of rows of this op in this chunk that
another air already collected, so the collector must ignore them. It was applied inside the
`while rows > 0` loop, i.e. once per instance the batch spans.

When a `(chunk, op)` batch crosses an instance boundary, the second iteration first calls
`count_to_skip()`, which already moves the rows consumed by the previous instance into
`skip_*_rows`, and then adds `skip` a second time. With `skip = 5` and a 15-row batch over
10-row instances, the second instance gets `initial_skip = 20` instead of `15`.

Those 5 rows are then collected by nobody: the first instance had already passed them and the
second skips over them. In debug builds the collector's `debug_assert!(is_final_skip())` fires;
in release the witness is silently short of inputs and the proof fails on the bus.

Reachable in the `Dma`/`DmaPrePost` paths, which are the only ones passing `skip > 0` — the
remainder-packing routes of bug 1 — for the single chunk where the remainder budget is partially
consumed, when that chunk's remaining rows still exceed the space left in the instance.

Fixed by applying the skip once per batch. The four near-identical match arms were extracted into
`add_skip_rows` / `add_count_rows`, which is what was hiding the bug: with the per-batch skip and
the per-instance count interleaved in one `match`, nothing signalled that the two have different
scope.

## 3. `inputs_counter` is cumulative but consumed as a per-chunk value

`DmaInstancesBuilder::inputs_counter` was never reset — not by `count_to_skip`, nor
`reset_count_and_skip`, nor `open_new_instance`, nor `flush`. It is stored per
`(instance, chunk)` record and read back by the collectors as the number of inputs *for that
chunk*: `Vec::with_capacity(num_inputs)` and the early-exit condition
`inputs.len() == num_inputs`.

So every record after the first carried the running total of the whole execution. Consequences:

- **Memory.** `Vec::with_capacity` over-allocates in proportion to the total run, not the chunk,
  at 32 B per `DmaInput`, once per collector and there is one per (instance, chunk). On a run
  with millions of DMA operations this is hundreds of MB of dead capacity.
- **Dead checks.** The collectors' early exit never triggers, so the
  `debug_assert!(is_final_skip())` guarding it never runs.

The witness itself stayed correct: the real filter is `collect_counters`, and `num_inputs` only
gates early termination.

Fixed by restarting the counter in `flush_current_chunk`, right after the record that owns it is
stored. Doing it there rather than in the callers means the invariant does not depend on all
three flush sites remembering to. The early-return path needs no reset: `inputs_counter` only
grows in the same loop iteration as one of the `count_*` counters, so if all of them are zero it
is zero too.

### Note on the per-instance value

`inputs_counter += inputs` is still charged once per loop iteration, i.e. a batch that straddles
an instance boundary charges its full `inputs` to both instances. That is deliberate and now
documented in the code.

`num_inputs` only controls the collectors' early exit, so **over-estimating is safe and
under-estimating would truncate collection**. An operation crossing an instance boundary is
collected — partially — by both instances, and the aggregate `(rows, inputs)` counters carry no
information about how the operations are distributed inside the batch, so an exact split is not
derivable here. Charging `inputs` to every instance the batch reaches is the correct upper bound;
it is tight whenever nothing straddles, which is the common case.

With the reset in place, every record now holds what only the first record held before this PR.

## Tests

10 unit tests, the first in this crate, following the `#[cfg(test)] #[path = …] mod tests;`
layout already used by `precomp-arith-eq`:

- `src/tests/dma_strategy_tests.rs` — instance counts for `Dma`/`DmaPrePost`, including an
  `assert_fits` helper that checks the invariant `DmaInstancesBuilder` panics on (the rows routed
  to each air fit in the instances requested, and no row is routed twice), and a sweep over values
  just below, at and just above each multiple of the capacity.
- `src/tests/dma_instances_builder_tests.rs` — per-(instance, chunk) bookkeeping: skip across an
  instance boundary, and the scope of the inputs counter across chunks, instances and ops.

Each fix was reverted individually to confirm its tests fail without it:

| reverted fix | failure |
|---|---|
| 1 | `full air overflows: 150 rows in 1 instances` — exactly the condition `open_new_instance` aborts on. The boundary sweep also caught a second case I had not anticipated (`memcmp = 100, inputcpy = 1`). |
| 2 | `initial_skip` 20, expected 15 |
| 3 | `num_inputs` 3, expected 1; and 4, expected 1 |

## Verification

- `cargo test -p precomp-dma` — 10/10 pass
- `cargo clippy -p precomp-dma --all-targets` — clean
- `cargo fmt -p precomp-dma -- --check` — clean
- `cargo check -p executor --all-targets` — clean (`executor` is the only crate depending on
  `precomp-dma`)

### DIFF
--- a/precompiles/dma/src/dma_instances_builder.rs
+++ b/precompiles/dma/src/dma_instances_builder.rs
@@ -122,6 +122,10 @@ impl DmaInstancesBuilder {
                 .chunks
                 .insert(chunk_id, (self.inputs_counter as u64, collect_counters));
             self.instances.last_mut().unwrap().last_chunk = Some(chunk_id);
+            // The inputs counter belongs to the (instance, chunk) record just stored: the collector
+            // built from it uses the value as its own bound. Restart it so the next record does not
+            // inherit this one's inputs.
+            self.inputs_counter = 0;
         }
     }
     #[inline(always)]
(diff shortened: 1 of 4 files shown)


rickb80 (approved):
LGTM

PR #1252, opened 2026-08-11
zisk-contracts: sync verifier constants with rebuilt snark setup

RogerTaule:
The checked-in Solidity verifiers were generated against an older provingKeySnark, so the Hardhat test reverted against a proof wrapped with the current setup. Regenerating the setup writes fresh contracts to build/provingKeySnark/final/ but does not update these copies.

- PlonkVerifier.sol: Qcx/Qcy now match final.zkey. These are PLONK VK constants, so a stale pair fails the pairing check and surfaces only as a bare "Expected transaction NOT to be reverted" with no indication of which constant is wrong.
- ZiskVerifier.sol: getRootCVadcopFinal() now matches provingKey/zisk/vadcop_final/vadcop_final.verkey.json. The test checks this off-chain before the call, so a mismatch here already reported itself clearly.
- test/plonk_proof.bin: re-wrapped against the same setup.

PlonkVerifier.sol is now byte-identical to the generated build/provingKeySnark/final/PlonkVerifier.sol. ZiskVerifier.sol is deliberately not a straight copy: the generated file reports its version as v1.0.0-beta, and the v1.1.0-alpha string here is kept.

Verified with:
  cargo-zisk-dev export-solidity-calldata -p test/plonk_proof.bin \
    -o fixtures/fixture.json
  ZISK_SOLIDITY_FIXTURE=fixtures/fixture.json npx hardhat test
  => 1 passing


rickb80 (approved):
LGTM

PR #1258, opened 2026-08-14
fix(mem): use aligned max addr in aligned address discard check

maximmenshikov:
Compare against aligned_max_addr instead of filter_max_addr in discard_align_addr, so aligned accesses are filtered by the checkpoint's aligned address range rather than the unaligned filter bound.


maximmenshikov:
Just noticed it is fixed in pre-develop-1.1.0-alpha

PR #1255, opened 2026-08-12
feat(keccakf): parametrize the AIR layout by `lanes_per_row`

hecmas:
## Motivation

The Keccakf AIR had a fixed trace shape: one row per round, the full 1600-bit state per row (~2,140 committed + ~880 bus columns, 14.3 GB of prover memory per 2^18 instance). This PR makes the width/height trade-off a template parameter, so the same constraint system can be instantiated anywhere between 25×1600 and 625×64 per keccak-f — without changing the arithmetization, the table, or the max constraint degree (3).

## Design

`Keccakf(N, lanes_per_row)` with `lanes_per_row` ∈ [1, 25]:

- **Layout**: each round-state occupies `ROWS_PER_STATE = ceil(25 / lanes_per_row)` consecutive rows; lane `l` lives entirely in group-row `l / lanes_per_row` at columns `[(l % lanes_per_row)·64, …+64)`. Lane-aligned splitting keeps every ρ rotation and every 64-bit memory word inside a single row. When `lanes_per_row` does not divide 25, the last group-row is ragged (its trailing columns stay zero); for a target R the densest choice is `lanes_per_row = ceil(25/R)`.
- **Round constraints**: the round function is enforced once per group-row. The anchor at group-row `k` rebuilds the full θ→ρ→π→χ→ι expression from a row-shifted view of its state-group and binds only the accumulators of its own lanes: `in_use_round[k] · (chunk_acc[i] − acc_val_k) === 0`. The five step functions are untouched.
- **Chunk lookups are group-row-invariant**: the tuple `[chunk_acc[i], state[…]'(ROWS_PER_STATE), …]` is identical at every anchor row, so a single `lookup_assumes` per chunk column (gated by the union selector `round_f_active`) serves all group-rows. This keeps the bus/im cost independent of R instead of multiplying with it.
- **Selectors**: a single fixed mask `ROUND_CLK_0` shifted by `k` selects group-row `k`; the per-`k` selectors are committed (`<==`, auto-computed) to keep bindings at degree 3. At R = 1 no extra column is committed and the selector degenerates to the previous `round_f_active`.
- **Memory interface**: `lanes_per_row` parallel ops over the first/last `ROWS_PER_STATE` clocks, with the word address advancing by `8·lanes_per_row` per clock (`clock_shift` `delta`). Ops beyond the ragged row's lane count skip it; the two window shapes are built once and shared so the STD materializes a single intermediate per shape.
[...]

PR #1260, opened 2026-08-15
feat(keccakf): χ-row S-box arithmetization

hecmas:
## Motivation

The previous Keccakf AIR computed each full round — θ, ρπ, χ, ι — as one degree-2 expression per output bit, with values growing up to 144, and then normalized arbitrary 3-bit chunks of those expressions through a base-145 lookup. That choice forced the two dominant costs of the AIR:

- **534 committed `chunk_acc` columns**, needed only to keep the degree-2 expressions out of the logup denominators (they would otherwise break im-column clustering), and
- **a 145³ ≈ 3.05M-row table** (2²² padded), because the value range 145 caps a lookup at 3 output bits.

Measured with the current toolchain, the old design costs **~75.5k committed cells per keccak-f** (2,137 stage-1 + ~880 stage-2 columns × 25 rows) — by far the most expensive precompile per operation.

## Design

χ is not an arbitrary function: it maps each 5-bit row (fixed y,z) to a 5-bit row using only that row, and everything before it in the round — θ, ρ, π — is 𝔽₂-linear. So the round is verified with **one lookup per χ-row**, keeping all committed expressions linear and moving the entire nonlinearity into the table:
```
( rc·12⁵ + Σₓ 12ˣ·b[x,y,z],  state'[0,y,z], …, state'[4,y,z] )  ∈  KECCAKF_TABLE
```

- `b = (π∘ρ∘θ)(state)` are plain integer sums of state bits (values ≤ 11; only parities carry keccak semantics), so the packed input is a **degree-1 expression** — no committed accumulator needed.
- The table (2·12⁵ = **497,664 rows**) decodes the five parities, applies χ, and XORs the ι bit into the x = 0 output. ι enters as a table *input*: y = 0 χ-rows pass the clock-weighted constant `wrc[z]`, all others pass 0, so every trace row stores the true post-ι round state and the memory interface needs no correction.
- The five output entries are the committed next-state bits themselves, pinned entry-wise against the table — 320 lookups per row instead of 534, and `chunk_acc` disappears entirely.

Booleanity of the state columns is load-bearing: it makes the base-12 packing decode uniquely and, at the input state, keeps the memory packing injective (e.g. bits (3,0) and (1,1) pack to the same memory word but have different
column parities downstream).

The memory access map, bus interface and `proves_operation` integration are unchanged; no main-CPU or emulator changes.

## `lanes_per_row` layout parameter

The AIR is parametrized by `Keccakf(N, lanes_per_row)` (divisors of 25): each round-state occupies `25 / lanes_per_row` consecutive rows, trading width for height at a fixed arithmetization. Narrow layouts share the 320 per-round lookups across group-rows through a committed `sbox_acc[64]` (the packed S-box input, bound at each χ-row's anchor row), which is what the width reduction costs. Full matrix in `precompiles/keccakf/pil/NUMBERS.md`:

|                       | LPR=25 (this PR) | LPR=5           | LPR=1            |
|-----------------------|------------------|-----------------|------------------|
| Trace shape           | 25×1600          | 125×320         | 625×64           |
| Total cols            | 2,168            | 508             | 240              |
| Prover mem / instance | 10.58 GB         | 2.60 GB         | 1.08 GB          |
| Cells / keccakf       | 54,200           | 63,500 (+17.2%) | 150,000 (+177%)  |

This PR instantiates `lanes_per_row: 25`. A more descriptive explanation can be found in #1255.

## Numbers (measured, same toolchain, N=2¹⁸)

|                    | old (chunked)  | new (χ-row S-box) |         |
|--------------------|----------------|-------------------|---------|
| Stage-1 columns    | 2,137          | 1,603             | −25%    |
| Stage-2 columns    | ~880           | 559               | −36%    |
| **Cells / keccakf**| **~75,500**    | **54,200**        | **−28%**|
| Constraints        | ~2,750         | 1,816             | −34%    |
| Table rows         | 3,048,625      | 497,664           | 6.1×    |


hecmas:
Superseded by the more optimal #1264.

PR #1264, opened 2026-08-17
feat(keccakf): sliced χ-row S-box arithmetization

hecmas:
## Motivation

Keccak-f was our most expensive precompile: ~75.5k committed cells per permutation (2,137 stage-1 + ~880 bus columns at 25 rows/op), driven by two structural costs — one bit per witness cell for all 1,600 state bits of every round, and 534 `chunk_acc` columns whose only job was to linearize degree-2 round expressions for a 145³-row lookup table.

This PR replaces the arithmetization with the **χ-row S-box** design (see #1260) **plus 2-way instance slicing**: every round cell carries two independent permutations, `v = a + 8·b`. Result: **38.5k cells per keccak-f (−49%)** at the default layout, with the same memory map and bus interface.

## Design

**χ-row S-box.** Everything before χ is 𝔽₂-linear, so the committed expressions are plain integer sums (per-slot values ≤ 5); χ — the only nonlinear step — acts independently on each 5-bit row (y,z) and is delegated entirely to a lookup: `( rc·28⁵ + Σₓ 28ˣ·b[x,y,z], state'[0,y,z], …, state'[4,y,z] ) ∈ KECCAKF_CHI_TABLE`, where `b = π∘ρ∘θ(state)` is degree-1 and ι enters as a table *input* (y = 0 rows pass the clock-weighted `wrc[z]`, applied to both slots — the two ops of a slot share the round).

**Slicing needs a normalization layer.** With two instances per cell, feeding the dirty θ-range [0,11] to the χ table would need a ~2³⁶ table. Instead, θ's column sums (slot values in [0,5]) are normalized to committed sliced parities `c` through a tiny **xor5 table** (36³ rows, three positions per lookup); θ-outputs rebuilt from clean parities keep each slot in [0,3], making the χ table 2·16⁵ = 2²¹ rows.

**Slot layout (29 state-groups).** Round rows are sliced; the boundary groups hold each op's plain bits so memory packs per-instance words:
```
IN_A | IN_B | round0 … round24 (sliced, + c) | OUT_A | OUT_B
```

Linear glue ties them: `round0 = IN_A + 8·IN_B`, `round24 = OUT_A + 8·OUT_B`, with booleanity on the boundary groups (load-bearing: it makes the glue and the memory packing decode uniquely). Each op has its own `in_use` flag, `step_addr` slots, memory ops and `proves_operation`; `in_use_b ⇒ in_use_a` since the sliced rounds are verified under op A's flag. An odd trailing op pairs with a zero-state dummy whose bus/memory flags stay off.

**`lanes_per_row` parameter** (divisors of 25): every state-group spans `25/lanes_per_row` rows, trading width for height at a fixed arithmetization. Narrow layouts share the per-round lookups across group-rows: the packed χ-input is bound to a committed `sbox_acc[64]` by a masked union constraint, and the xor5 inputs become mask-gated degree-2 expressions (cheaper than committing the sums). A more descriptive explanation can be found in #1255.

## Numbers (measured, N = 2¹⁸)

|                        | old (main)  | **LPR=25**  | LPR=5           | LPR=1           |
|------------------------|-------------|-------------|-----------------|-----------------|
| Trace shape (per op)   | 25×~3,020   | 14.5×2,652  | 72.5×643        | 362.5×273       |
| Stage-1 / stage-2 cols | 2,137 / 880 | 1,925 / 721 | 453 / 184       | 146 / 121       |
| Constraints            | ~2,750      | **1,873**   | 458             | 177             |
| Ops per instance       | 10,485      | **18,078**  | 3,614           | 722             |
| **Cells / keccakf**    | **~75,500** | **38,454 (−49%)** | 46,617 (+21%) | 98,962      |
| Witness buffer / instance | ~6.3 GB  | ~5.6 GB     | **~1.35 GB**    | ~0.57 GB        |
| Table rows             | 3,048,625   | 2,097,152 + 46,656 (two virtual tables) |||

Supported operating points: **25** (cells) and **5** (memory); 1 compiles but is dominated (lookup-call floor × 725 rows).


RogerTaule (dismissed):
LGTM!

RogerTaule (approved):

PR #1291, opened 2026-08-30
Add BLAKE2s round precompile

amiabix:
Adds a BLAKE2s round precompile alongside the existing BLAKE2b implementation.

BLAKE2s uses 32-bit words, so its additions need a simpler decomposition than BLAKE2b. The two rotations that are not byte-aligned are handled as a byte rotation followed by a small shift:

```
>>> 12 = (<<< 16) <<< 4
>>>  7 = (<<< 24) <<< 1
```

The carry values used by those shifts are constrained to their unique valid values, rather than merely range-checked. In particular, the >>> 12 carry is bound through the packed vb_prime byte columns. This is important because >>> 12 occurs mid-round: unlike BLAKE2b's final non-byte-aligned rotation, it is not constrained by a transition into the next row.

The existing BLAKE2b XOR lookup table is reused unchanged because it operates byte-wise. BLAKE2s words are stored as one 32-bit value per 64-bit memory slot, with the upper half constrained to zero, preserving the existing memory schedule.

Round indices are rejected instead of reduced modulo 10. syscall_blake2s_round validates the full u64 index before narrowing or issuing the syscall; callers require a successful return. This avoids aliasing 2^32 to round zero and matches the AIR, where round_idx is a one-hot selector over the ten valid schedules and is tied to the memory-supplied index.

The operation-bus dispatch is now explicit through ZiskOperationType::uses_operation_bus() rather than relying on enum ordering. This covers all existing core operations and prevents a newly declared precompile from being silently omitted from the bus.

Also corrects the stale BLAKE2_COST constant. Its old formula predates the AIR and clock changes in 21d497cd3; both BLAKE2b and BLAKE2s costs now follow the compiled-AIR cost convention used by the other precompiles.


hecmas:
Superseded by #1300.

PR #1299, opened 2026-09-05
Fix PLONK wrap of a plain vadcop_final proof: use the vadcop_final verkey

sembrestels:
## Summary

`cargo-zisk wrap --plonk` (and the SDK's `wrap_proof(..., ProofKind::Plonk)`) fails on a plain `vadcop_final` proof, and would export a wrong `rootCVadcopFinal` if it succeeded. The single-process `cargo-zisk prove --plonk` path is unaffected.

## Bug

`ProverBackend::plonk` reads the leading `PROGRAM_VK_LEN` publics of the input proof and

1. passes them to `generate_final_snark_proof` as the RecursiveF verkey override, and
2. stores the same words as `rootc` in the resulting `Proof`.

Those publics are only a verkey for an aggregated (`Recurser`) proof. For a plain `vadcop_final` proof they are the program VK (`rom_root`). The comment on `generate_final_snark_proof` already describes the consequence:

> using vadcop_final's would diverge the wrapper transcript (VerifyPoW aborts)

and the same holds in reverse: stamping the program VK for a plain proof diverges the transcript.

Observed on `v1.2.0-alpha` (Linux x86_64, CPU build, keys installed by `ziskup`):

```
$ cargo-zisk prove -e guest.elf -i input.bin -k ~/.zisk/provingKey -o proof_stark -y   # ok, "Proof verified successfully"
$ cargo-zisk wrap -p proof_stark -k ~/.zisk/provingKey -w ~/.zisk/provingKeySnark --plonk -o proof_plonk
...
proofman::recursion INFO: >>> GENERATE_RECURSIVEF
Failed assert in template/function VerifyPoW line 26. Followed trace of components: main.sV.Transcript0_8111_726168.calculateFRIQueries0_558_32689.VerifyPoW_23_
```

With only the first half fixed (override = `None`), the wrap succeeds but `cargo-zisk-dev export-solidity-calldata` emits `rootCVadcopFinal == programVK`, and `ZiskVerifier.verifySnarkProof` reverts with `InvalidProof()` because the contract hashes `sha256(programVK ‖ publicValues ‖ rootCVadcopFinal)` and the circuit committed to the `vadcop_final` verkey.

## Fix

The verkey/rootC decision is extracted into a pure helper, `plonk_wrap_verkey(publics_full, vadcop_final_vk) -> (verkey_override, rootc)`, driven by `VadcopKind::from_publics_full`:

- `Recurser` → the proof's own leading verkey for both (unchanged behaviour for aggregated proofs),
- otherwise → `None`, i.e. the `vadcop_final` verkey, matching what `prove --plonk` does, and `rootc = vadcop_final verkey`.

## Tests

Two unit tests in `prover-backend/src/prover/backend.rs` build raw vadcop publics (`[flag | leading vk | inputs]`) and exercise the helper without a snark setup:

- `plain_vadcop_final_proof_is_checked_against_the_vadcop_final_verkey` (flag = 1): expects no override and `rootc == vadcop_final verkey`. On the previous behaviour it fails with `left: Some([program VK]) right: None`.
- `aggregated_proof_is_checked_against_its_own_verkey` (flag = 0): expects the leading verkey for both. Passes before and after.

```
cargo test -p zisk-prover-backend prover::backend::tests
```

## Verification

- `cargo check -p zisk-prover-backend` and the tests above pass.
- Rebuilt the SDK with this change and wrapped the saved `vadcop_final` proof of a small guest (fibonacci, 2544 steps): `GENERATE_RECURSIVEF` completes, a PLONK proof is produced, `export-solidity-calldata` now yields `rootCVadcopFinal = 0x564c2b1b…4f84382f` = `ZiskVerifier.getRootCVadcopFinal()`, and a Foundry test calling `verifySnarkProof` with the exported fields passes (about 360k gas); flipping one byte of `publicValues` reverts with `InvalidProof()`.
- Not tested end to end: wrapping an aggregated (`Recurser`) proof. That branch keeps the previous behaviour and is covered by the second unit test only.

The same code is present on `main`.

### DIFF
--- a/prover-backend/src/prover/backend.rs
+++ b/prover-backend/src/prover/backend.rs
@@ -531,14 +531,21 @@ impl ProverBackend {
         let vadcop_final_proof =
             VadcopFinalProof::new(proof.to_vec(), publics_full.to_vec(), false, self.hash()?);
 
-        // Read the program VK from the flag-free view (a full vadcop_final
-        // publics vector carries the `is_vadcop_final_proof` flag at index 0).
-        let proof_verkey = &program_publics(publics_full)[..PROGRAM_VK_LEN];
+        // The RecursiveF verifier checks the vadcop proof against the verkey stamped
+        // into its leading publics. Only an aggregated (Recurser) proof carries its
+        // own verkey there; a plain vadcop_final proof carries the program VK
+        // (rom_root), and stamping that as the verkey diverges the wrapper
+        // transcript (circom `VerifyPoW` assert). Read the flag-free view (a full
+        // vadcop_final publics vector carries `is_vadcop_final_proof` at index 0).
+        let verkey_override = match VadcopKind::from_publics_full(publics_full) {
+            VadcopKind::Recurser => Some(&program_publics(publics_full)[..PROGRAM_VK_LEN]),
+            _ => None,
+        };
         let snark_proof = self
             .snark_wrapper
             .as_ref()
             .unwrap()
-            .generate_final_snark_proof(&vadcop_final_proof, Some(proof_verkey))?;
+            .generate_final_snark_proof(&vadcop_final_proof, verkey_override)?;
 
         let time = start.elapsed();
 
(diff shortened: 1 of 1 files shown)


sembrestels:
CLA: signed via 0xPolygonHermez/clabot-config#181. Once that is merged, the CLA check here should pass on its next run.

PR #1300, opened 2026-09-07
feat(precompiles): adding the Blake2s precompile

hecmas:
## Summary

Adds a **BLAKE2s** precompile that proves the full 10-round permutation of the compression function in a single syscall, reusing the Blake3f arithmetization. Along the way the byte-wise XOR⊕ROTR lookup table becomes shared by the three Blake AIRs, and the existing Blake2 precompile is renamed **Blake2b** everywhere so the two variants are distinguishable.

The BLAKE2s round function is BLAKE3's G (32-bit words, rotations 16/12/8/7, same column and diagonal indexing) with the RFC 7693 message schedule. `blake2s.pil` is therefore `blake3f.pil` with `NUM_ROUNDS = 10`, the 10-row SIGMA table, and its own opcode and lookup id; the per-row constraints, transition constraints and rotation tricks are identical.

## PIL

- `precompiles/blake2/pil/blake2s.pil` (new): `Blake2s` airtemplate, 80-row cycle (10 rounds × 8 G rows), 51 stage-1 columns, packed memory layout (two u32 per u64 word, the natural layout of a `[u32; 16]`), single memory port with 26 memory operations per permutation. Message consistency across rounds via the `BLAKE2S_MSG_ID` lookup keyed by the main step, as in Blake3f. Instantiated as `Blake2s(N: 2**18)` → 3276 permutations per instance.
- `precompiles/blake3/pil/blake_table.pil` (renamed from `blake3f_table.pil`): `BlakeTable`, `BLAKE_TABLE_ID = 129`, 2^17 rows. Rows with `rot = 0` are the plain byte XOR, rows with `rot = 12` serve the mid-G rotation of Blake2s and Blake3f. Blake2br now uses this table with `rot = 0`; `blake2br_table.pil` (2^16 rows) is deleted. All three AIRs require it with the same literal path (`require "blake3/pil/blake_table.pil"`), which the compiler needs for deduplication.
- `blake3f.pil`: no semantic change; clock sets and write clocks expressed via `NUM_ROUNDS` / `WRITE_STATE_START` so the two files diff cleanly.
- `OP_BLAKE2S = 0xEF`, `BLAKE2S_MSG_ID = 131`; `zisk.pil` instantiates `Blake2s` after `Blake3f` (air id 29, the Dma/JumpDest/virtual-table ids shift by one) and one `virtual BlakeTable`.

Compared with a per-round BLAKE2s design (as in #1291), one compression costs the same 80 trace rows but 26 memory operations instead of 510 and 1 syscall instead of 10.

Supersedes the per-round approach of #1291.


RogerTaule (approved):

zkronos73 (approved):
LGTM

PR #1304, opened 2026-09-08
fix soundness bug with opcode mapping, normalize opcodes.

zkronos73:
The `Binary` air proves an opcode **and** its 32-bit variant at `opcode + 0x10` from the same row,
because `mode32` is a free witness. Every opcode handed to `Binary` therefore silently reserves
`opcode + 0x10` too. Four opcodes in the bit-manipulation block violated that: their shadow slots
had been given to `BinaryExtension` operations, so `Binary` could satisfy those operations on the
operation bus with the wrong semantics.

This PR reorganizes the opcode space so no shadow slot is ever shared, and adds a unit test that
fails if a future opcode reopens the hole. It also removes the duplicated opcode constants that had
accumulated across the tree and aligns the PIL constant names with `ZiskOp`.

Nothing in the instruction set changes semantically — only numeric opcode assignments, plus the
fixed-column data that depends on them.

---

## The reference: `core/src/zisk_ops.rs`

`define_ops!` in [`core/src/zisk_ops.rs`](../core/src/zisk_ops.rs) is the **single source of truth**
for the ZisK instruction set. One row per operation:

```rust
(Variant, "name", OpType, COST, code, input_size, output_size, opc_fn, op_fn, stats_fn)
```

`OpType` is what decides which secondary state machine has to prove the operation (`Binary`,
`BinaryE`, `ArithAm32`, `PubOut`, …), and `code` is the opcode that travels on the operation bus.

### The generated constants — use these, do not redefine them

For every row the macro emits an associated constant on `ZiskOp` whose name is the **uppercase of
the string name**, not of the enum variant:

| row | enum variant | generated constant |
|---|---|---|
| `(Add, "add", …, 0x0a, …)` | `ZiskOp::Add` | `ZiskOp::ADD` |
| `(SignExtendB, "signextend_b", …, 0x27, …)` | `ZiskOp::SignExtendB` | `ZiskOp::SIGNEXTEND_B` |
| `(OrcB, "orc_b", …, 0x3e, …)` | `ZiskOp::OrcB` | `ZiskOp::ORC_B` |
| `(SllUW, "sll_u_w", …, 0x43, …)` | `ZiskOp::SllUW` | `ZiskOp::SLL_U_W` |

Each one is a plain `const u8`, so it works in every position an opcode is needed:

```rust
const PADDING: u8 = ZiskOp::ADD;                    // const position
if inst.op == ZiskOp::FCALL { … }                   // comparison
match op {                                          // pattern position
    ZiskOp::BREV8 => …,
    ZiskOp::ANDN  => …,
    _ => …,
}
enum TableOp { Andn = ZiskOp::ANDN as u16, … }      // enum discriminant
```

Pattern position is the one that matters in practice: `ZiskOp::Andn.code()` is a function call and
cannot be a pattern, which is why the tree had grown ~168 local aliases of the form
`const ANDN_OP: u8 = ZiskOp::Andn.code();` and, in `arith_eq_row.rs`, a `match` built entirely out
of `x if x == …` guards. All of them are gone; reference `ZiskOp::<NAME>` directly.

Also on `ZiskOp`, useful when iterating: `try_from_code()`, `name()`, `op_type()`, `cost()`,
`MIN_OPCODE`, `MAX_OPCODE`, `OPCODES_COUNT`.

### The PIL mirror

[`pil/operations.pil`](../pil/operations.pil) declares the same opcodes as `const int OP_<NAME>`,
using the **same** uppercase-of-the-string-name rule. PIL and Rust are now 1:1 by name, so
`OP_SIGNEXTEND_B` ↔ `ZiskOp::SIGNEXTEND_B`, with no alias table needed to check one against the
other. Keep it that way when adding an opcode: add the `define_ops!` row first, then mirror it.

### The four exceptions

Four opcodes that `Binary` proves are **not** ZisK instructions — the transpiler never emits them
and they never appear in a ROM. `Arith` puts them on the operation bus to check
`0 <= |remainder| < |divisor|` after a division. They have no `ZiskOp` variant and are declared in
[`state-machines/binary/src/binary_constants.rs`](../state-machines/binary/src/binary_constants.rs)
(and in `operations.pil`):

| const | code | role |
|---|---|---|
| `GT_OP` | `0x08` | signed `>`; used when remainder and divisor are both negative, where `\|d\| < \|b\|` is exactly `d > b` signed |
| `GTW_OP` | `0x18` | m32 shadow of `GT_OP`; nothing emits or proves it, it exists to record that the slot is taken |
| `LT_ABS_NP_OP` | `0x50` | `\|a\| < \|b\|` for `a < 0`, `b > 0` |
| `LT_ABS_PN_OP` | `0x51` | `\|a\| < \|b\|` for `a > 0`, `b < 0` |

That file holds nothing else: everything that has a `ZiskOp` variant must be read from `ZiskOp`.

---

## The feature: `Binary` proves both an opcode and its `m32` variant

`Binary` is a byte-sliced air. A 32-bit operation is the same byte chain as its 64-bit counterpart,
stopping at byte 3 and sign-extending the top half, so one air covers both. The mode is a column,
and the opcode put on the bus is derived from it —
[`state-machines/binary/pil/binary.pil`](../state-machines/binary/pil/binary.pil):

```
col witness bits(7) b_op;      // binary operation code
col witness bits(1) mode32;    // 1 if 32-bit mode, 0 if 64-bit

proves_operation(op: b_op + 0x10 * mode32, a:, b:, c:, flag: cout);
```

That is why every 32-bit opcode is exactly its 64-bit opcode plus `0x10`: `minu` `0x02` /
`minu_w` `0x12`, `add` `0x0a` / `add_w` `0x1a`, and so on. `M32_OFFSET` in `binary_constants.rs`
is the Rust-side name for that `0x10`.

**The part that bites.** `mode32` is a free witness. The only thing constraining it is
`mode32 * (1 - mode32) === 0`, and `b_op` is constrained by the byte-table lookup — but *nothing*
ties `mode32` to the `m32` flag of the main air. The prover chooses it. So for any `b_op` the air
can legitimately produce a row for `b_op` **or** for `b_op + 0x10`.

The operation bus is a lookup: main *assumes* `[op, a, b, c, flag, …]` and the state machines
*prove* it. Consequently:

> Every opcode assigned to `OpType::Binary` reserves `opcode + 0x10` as well. That slot must hold
> either the operation's own `m32` variant, or nothing.

Give the slot to an unrelated operation and `Binary` becomes able to prove that operation's bus
tuple with the wrong result.

---

## The bug

Before this PR, the bit-manipulation opcodes proved by `Binary` sat at `0x32`-`0x35`, in the middle
of the `BinaryExtension` block — so their shadows landed on live `BinaryExtension` opcodes:

| `Binary` op | code | shadow | slot was | proved by |
|---|---|---|---|---|
| `brev8` | `0x32` | `0x42` | `cpop_w` | `BinaryExtension` |
| `andn` | `0x33` | `0x43` | `orc_b` | `BinaryExtension` |
| `orn` | `0x34` | `0x44` | `bclr` | `BinaryExtension` |
| `xnor` | `0x35` | `0x45` | `bext` | `BinaryExtension` |

`sh1add`/`sh2add`/`sh3add` (`0x49`/`0x4b`/`0x4d`) happened to shadow free slots, and the base ops
all had genuine `_w` variants, so those were fine.

**Exploit shape.** Take a `cpop_w` instruction (`0x42`). The honest prover satisfies main's
assumption with a `BinaryExtension` row. A malicious prover instead sets `b_op = 0x32` (`brev8`),
`mode32 = 1` in `Binary`, which proves the tuple `[0x42, a, b, c', …]` where `c'` is the byte-reverse
result rather than the population count. The bus balances, the proof verifies, and a `cpop_w`
instruction has returned an arbitrary attacker-chosen value. The same applies to `orc_b`, `bclr`
and `bext`.

**Why it went unnoticed.** The rule is invisible while every `Binary` opcode has a real `_w`
counterpart, which was true of the original instruction set. The B-extension opcodes were added
later and placed by family rather than by proving air, so the `Binary`-proved ones ended up
interleaved with the `BinaryExtension` block.

---

## The fix

`Binary`-proved bit manipulation moved out of the `BinaryExtension` block into its own region,
whose shadows (`0x62`-`0x68`) are reserved and empty:

```
Binary                              m32 shadow
  0x02-0x10  base 64-bit ops    ->  0x12-0x20  the real _w variants
  0x50       lt_abs_np          ->  0x60       reserved
  0x51       lt_abs_pn          ->  0x61       reserved
  0x52-0x55  brev8 andn orn xnor->  0x62-0x65  reserved, must stay empty
  0x56-0x58  sh1add sh2add sh3add-> 0x66-0x68  reserved, must stay empty
  0x59-0x5f  free               ->  0x69-0x6f  reserved

BinaryExtension (no shadow rule: it proves `op` directly)
  0x21-0x29  sll..sext_w
  0x30       rev8
  0x31-0x33  pack pack_h pack_w
  0x34-0x37  rol rol_w ror ror_w
  0x38-0x3b  clz clz_w ctz ctz_w
  0x3c-0x3d  cpop cpop_w
  0x3e       orc_b
  0x3f-0x42  bclr bext binv bset
  0x43       sll_u_w

"Software" (not proved by any air)
  0x90-0x93  add_u_w sh1/2/3add_u_w   (transpiler decomposes them; the opcode only carries the
  0x95-0x9b  clmul* xperm* czero*      cost of the resulting sequence, or is behind a *_native
                                       transpiler feature with no table rows yet)
  0x94, 0x9c-0x9f  free

  0xa0       pubout                    (was 0x30, which rev8 now uses)
```

Two placements worth calling out:

- **`lt_abs_np`/`lt_abs_pn` stay at `0x50`/`0x51`.** The first draft of the remap put `brev8`/`andn`
  there, which collided head-on with these two internal comparisons — same opcode in the `col fixed
  OP` of `binary_table.pil`, giving the row-layout `switch` duplicate cases. The block was shifted
  to `0x52` instead.
- **`sll_u_w` is at `0x43`, not in the `0x90` block.** Unlike the other `*_u_w` opcodes it *is*
  proved natively (it has a row block in `binary_extension_table.pil` and `zba_native` is on by
  default), so it belongs next to `bset`. Keeping it at `0x94` would also have forced
  `col witness bits(7) op` in `binary_extension.pil` up to 8 bits.

### Regression test

`no_binary_opcode_shadows_an_unrelated_operation` in `binary_constants.rs` walks every opcode
`Binary` can prove — the `OpType::Binary` rows of `define_ops!` plus the four internal ones — and
asserts that `opcode + 0x10` is either that operation's own `m32` variant or unassigned. Adding a
`Binary` opcode whose shadow is taken now fails the test with, e.g.:

```
0x10 (xor) shadows 0x20 (brev8): the m32 slot of a binary opcode must stay empty
```

---

## Also in this PR

- **Duplicated opcode constants removed**: all 279 uses of `ZiskOp::<Variant>.code()` are gone —
  168 of them were alias definitions (`const X_OP: u8 = ZiskOp::X.code();`) and the other 111 were
  call sites — replaced by the generated `ZiskOp::<NAME>` constants. The `frops-analyzer` generator
  emits the new form too, so regenerating the `@generated` frops files does not undo it.
- **PIL names aligned with `ZiskOp`**: 26 constants renamed (`OP_KECCAKF` → `OP_KECCAK`,
  `OP_EC_ADD_SECP256K1` → `OP_SECP256K1_ADD`, `OP_SEXT_B` → `OP_SIGNEXTEND_B`, …). Name-only change;
  the recompiled `PILOUT_HASH` is byte-identical across the rename.

## Notes for reviewers

- **Fixed data must be regenerated**: the frops tables are indexed by opcode
  (`OP_TABLE_OFFSETS[op - OP_TABLE_OFFSETS_START]`) and the `*_frops_fixed.bin` files carry the
  opcode column. `./tools/test-env/setup_build.sh --compile-pil` regenerates them plus
  `pil/zisk.pilout`. Already run: the only source change it produces is `PILOUT_HASH` in
  `pil/src/pil_helpers/traces.rs`, which confirms no air structure moved — only fixed-column
  contents.
- **When adding an opcode**: put the row in `define_ops!`, mirror the constant in
  `operations.pil` with the same name, and — if its `OpType` is `Binary` — leave `opcode + 0x10`
  free. The test enforces the last part.


fractasy (dismissed):

hecmas (dismissed):
LGTM

hecmas (dismissed):
LGTM

